PR #80, opened 2016-07-08
Make IDBRequest.error and IDBRequest.transaction nullable

nox:
The prose says that this field can be null.


inexorabletash:
Good catch, thanks!

Issue #74, opened 2016-04-18
deleteDatabase error case should never happen


inexorabletash:
The steps for `deleteDatabase` start off with

> 1. Let request be a new open request.
> 2. Queue a task to run these substeps:
>    1. Let result be the result of running the _steps for deleting a database_, with the origin of the IDBEnvironment used to access this IDBFactory, name, and request.
>    2. If result is an error set the error of request to result and dispatch an event at request. The event must use the Event interface and set the type attribute to "error". The event does bubble but is not cancelable.

But the "steps for deleting a database" never return an error. This may be legacy from timeout and/or sync behavior?

sicking:
Possibly. The only reason I can think of is file IO error.

I have no other recollection.

nolanlawson:
Is this just an implementation bug that this actually does throw? I'll double-check, but I'm pretty sure some engines (Edge iirc) throw if you rapidly open/delete databases with the same name several times.

inexorabletash:
We may want to keep this in the `deleteDatabase()` steps, but add something into "steps for deleting a database" to allow implementations to generate errors.

This is similar to the "forced close" addition; in the abstract it should never be needed, but in the real world implementations run into problems and how the errors appear to scripts should be defined.

PR #97, opened 2016-10-12
Define how implementations should fail open/deleteDatabase. For #74

inexorabletash:
@aliams, @bevis-tseng, @nolanlawson - anyone want to review this?

I added a simple "hook" in the algorithms for to open/delete to define where implementations that hit e.g. an I/O error on creation/deletion can fail. This is already handled in the API entry points (see #74) but not in the algorithms.


aliams:
This change looks reasonable to me. Perhaps we can go with something even less specific as a catch-all. Maybe something along the lines of: "if this fails for any reason If this fails for any reason, return an appropriate error..."

PR #100, opened 2016-10-20
Replace abrupt completion asserts/returns with !? shortcuts from ES spec

inexorabletash:
Calling an abstract ECMAScript operation with ! prefix is equivalent to asserting it will never return an abrupt completion. Calling an abstract ECMAScript operation with ? prefix is equivalent to following the call with ReturnIfAbrupt(result).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5681,7 +5681,7 @@ steps may throw an exception.
         4. Let |index| be 0.
         5. While |index| is less than |len|, run these substeps:
 
-            1. Let |hop| be [=!=] HasOwnProperty(|input|, |index|).
+            1. Let |hop| be [=?=] HasOwnProperty(|input|, |index|).
 
             2. If |hop| is false, return invalid.
 
--- a/index.bs
+++ b/index.bs
@@ -54,6 +54,9 @@ spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
         text: ArrayBuffer; url: #sec-arraybuffer-objects
         text: Uint8Array; url: #sec-typedarray-objects
         text: Promise; url: #sec-promise-objects
+        url: sec-algorithm-conventions
+            text: !
+            text: ?
 spec: webidl; urlPrefix: https://heycam.github.io/webidl/
     type: dfn
         text: sequence<DOMString>; url: #idl-sequence
@@ -5444,11 +5447,11 @@ ECMAScript value or failure, or the steps may throw an exception.
       <dd>
         1. If Type(|value|) is not Object, return failure.
 
-        2. Let |hop| be ! HasOwnProperty(|value|, |identifier|)
+        2. Let |hop| be [=!=] HasOwnProperty(|value|, |identifier|)
 
         3. If |hop| is false, return failure.
 
-        4. Let |value| be ! |value|.[<span>[</span>Get]](|identifier|,
+        4. Let |value| be [=!=] |value|.[<span>[</span>Get]](|identifier|,
             |value|)
 
         5. If |value| is Undefined, return failure.
@@ -5493,7 +5496,7 @@ as follows. The algorithm takes a |value|, a |key| and a |keyPath|.
         object (see [=structured clone algorithm=] [[!HTML]]), then
         [=throw=] a {{DataError}}.
 
-    2. Let |hop| be ! HasOwnProperty(|value|, |identifier|)
+    2. Let |hop| be [=!=] HasOwnProperty(|value|, |identifier|)
 
     3. If |hop| is false, run these substeps:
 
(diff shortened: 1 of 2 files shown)


inexorabletash:
@Domenic - could you review this?

(There more linking-to-ES that can be done. And at some point I'll pester you to review the bindings section of this spec in general to see if we can clean it up at all.)

index.bs
```diff
@@ -5444,18 +5447,14 @@ ECMAScript value or failure, or the steps may throw an exception.
...
+        3. If |hop| is false, return failure.
 
-        5. Let |value| be |value|.[<span>[</span>Get]](|identifier|,
+        4. Let |value| be [=!=] |value|.[<span>[</span>Get]](|identifier|,
             |value|)
 
-        6. Assert: |value| is not an abrupt completion.
-
-        7. If |value| is Undefined, return failure.
+        5. If |value| is Undefined, return failure.
```
domenic: lowercase undefined

index.bs
```diff
@@ -5444,18 +5447,14 @@ ECMAScript value or failure, or the steps may throw an exception.
       <dd>
         1. If Type(|value|) is not Object, return failure.
 
-        2. Let |hop| be HasOwnProperty(|value|, |identifier|)
-
-        3. Assert: |hop| is not an abrupt completion.
+        2. Let |hop| be [=!=] HasOwnProperty(|value|, |identifier|)
```
domenic: Why can't this fail? Does some earlier step prevent `value` from being a proxy?

index.bs
```diff
@@ -5444,18 +5447,14 @@ ECMAScript value or failure, or the steps may throw an exception.
...
 
-        4. If |hop| is false, return failure.
+        3. If |hop| is false, return failure.
 
-        5. Let |value| be |value|.[<span>[</span>Get]](|identifier|,
+        4. Let |value| be [=!=] |value|.[<span>[</span>Get]](|identifier|,
```
domenic: Same question, but here you'd need to prevent `value` from having any getters and having anything with a potential getter in its prototype chain.

index.bs
```diff
@@ -5679,39 +5674,34 @@ steps may throw an exception.
       <dt>If IsArray(|input|)</dt>
       <dd>
 
-        1. Let |len| be the ToLength(Get(|input|,
+        1. Let |len| be [=!=] ToLength(Get(|input|,
```
domenic: Need ? or ! for the Get, not just the ToLength

Is there something guaranteeing that ToLength will succeed? What about this?

```js
{
  get length() {
    return { valueOf() { throw new Error("boo"); } };
  }
}
```

index.bs
```diff
@@ -5679,39 +5674,34 @@ steps may throw an exception.
...
-            8. If |key| is invalid or an exception, then abort these
-                steps and return |key|.
+            6. If |key| is invalid abort these steps and return
+                invalid.
 
-            9. Append |key| to |keys|
+            7. Append |key| to |keys|
 
-            10. Increase |index| by 1
+            8. Increase |index| by 1
```
domenic: Nit: these steps are inconsistent about whether they end with a period or not. (They should, generally.)

index.bs
inexorabletash: There's more context in the spec (a note below the algorithm); the value being processed here is always the result of a structured clone operation, so there must be no proxies, getters, etc.

index.bs
inexorabletash: Ditto

index.bs
inexorabletash: done

index.bs
inexorabletash: Thanks re: needing ! for the Get.

This is in "if IsArray(input)" clause, i.e. we're dealing with an array exotic object. Is that sufficient?

domenic: Nah, consider the following:

```js
class EvilArray extends Array {
  get length() { /* as before */ }
}
```

Such objects will pass IsArray.

inexorabletash:
As more context, the "Extract a key from a value" and "Inject a key into a value" algorithms operate on clones, so are side-effect-free. (And necessarily so as some implementations may do operations asynchronously, computing index keys off thread, or deferring key injection until a value is retrieved, etc.)

index.bs
inexorabletash: Fun. I think implementations are likely to bypass the getter here for exotic array objects and access the internal length directly (Chrome does!).

Switched these to ? for the cases where the input is not a clone for now. 

(I local testing I don't see the length override getting called even when I invoke length on an instance. Hrm.)

inexorabletash: Ah, yes - an Array exotic's length is an own property, so the prototype's getter is ignored when getting length. Still, I bet proxies do bad stuff, so ? is correct here anyway but it'll be interesting to see if any impl trips up.

PR #101, opened 2016-10-20
Linkify more ES terms

inexorabletash:
* Use Get() rather than invoking [[Get]]
* Rely on IDL for sequence -> array conversions in operations
* Add missing periods at the end of sentences.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -23,46 +23,56 @@ Test Suite: https://github.com/w3c/web-platform-tests/tree/master/IndexedDB
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: dom.html
         type: interface
-            text: Document; url: #document
+            text: Document; url: document
     urlPrefix: browsers.html
         type: dfn
-            text: origin; url: #origin
+            text: origin; url: origin
     urlPrefix: workers.html
         type: dfn
-            text: worker; url: #worker
+            text: worker; url: worker
     urlPrefix: infrastructure.html
         type: dfn
-            text: strictly splitting the string; url: #strictly-split-a-string
-            text: structured clone; url: #structuredclone
-            text: structured clone algorithm; url: #safe-passing-of-structured-data
+            text: strictly splitting the string; url: strictly-split-a-string
+            text: structured clone; url: structuredclone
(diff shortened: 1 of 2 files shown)


inexorabletash:
@domenic - could you take a look at this follow-up?

index.bs
```diff
@@ -5390,22 +5383,31 @@ ECMAScript value or failure, or the steps may throw an exception.
...
 
         3. If |key| is failure, abort the overall algorithm and return
             failure.
 
-        4. Append the result of the first sub-step to end of |result|.
+        4. Let |p| be ! [=ToString=](|i|).
+
+	    5. Let |status| be [=CreateDataProperty=](|result|, |p|, |key|).
+
+		6. Assert: |status| is true.
```
domenic: Some mixed tabs and spaces here it seems

index.bs
```diff
@@ -5421,11 +5423,12 @@ ECMAScript value or failure, or the steps may throw an exception.
...
 
     <dl class=switch>
-      <dt>If Type(|value|) is String, and |identifier| is "<code>length</code>"</dt>
+      <dt>If [=/Type=](|value|) is String, and |identifier| is "<code>length</code>"</dt>
       <dd>Let |value| be a Number equal to the number of elements in |value|.</dd>
 
       <dt>If |value| is an [=Array=] and |identifier| is "<code>length</code>"</dt>
-      <dd>Let |value| be [=!=] ToLength([=!=] Get(|value|, "<code>length</code>")).</dd>
+
+      <dd>Let |value| be [=!=] [=ToLength=]([=!=] Get(|value|, "<code>length</code>")).</dd>
```
domenic: Link Get()?

index.bs
```diff
@@ -23,46 +23,56 @@ Test Suite: https://github.com/w3c/web-platform-tests/tree/master/IndexedDB
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
```
domenic: I wonder how many of these are still necessary.

domenic: Yeah I'm fixing the plural ones. Usually you should be able to use for="" or spec="" or lt="" to resolve the rest without using link-defaults.

Worker is an interface, not a term, so that's probably why it's confused there.

PR #103, opened 2016-10-31
Change "index handle" -> "object store handle" references within `IDB…

brettz9:
…ObjectStore.getKey`

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2695,8 +2695,8 @@ invoked, must run these steps:
 
 <div class=algorithm>
 
-1. Let |transaction| be this [=index handle=]'s
-    [=index-handle/transaction=].
+1. Let |transaction| be this [=object store handle=]'s
+    [=object-store-handle/transaction=].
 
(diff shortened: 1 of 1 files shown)


inexorabletash:
Apologies - this repo should have automatic IPR checks set up with it. Working on it...

This change would be considered non-substantive, but I want to get the check in place first.

Thanks for pointing out the issue, @brettz9 !

brettz9:
Sure... Let me know if there's already something I can sign (or I can wait for checks) in case any of my PRs do become substantive (at this point at least, I doubt it).

inexorabletash:
Applied a similar change in 57069af90471b544807d282ef67585c1bd1f598f

For tracking purposes, going forward can you file issues instead of or in addition to the PRs? Often it will be easier to make the change directly and I'd prefer to have an issue to link to so others can understand the problem.

inexorabletash:
And thanks again!

brettz9:
Sure (will try to remember)

PR #104, opened 2016-10-31
a unique direction is not possible by the time these steps are reached

brettz9:
…ObjectStore.getKey`


inexorabletash:
Applied similar changes in 58d9d4f147f02f567f09551c83de0d290d3c0913 - thanks for the report!

PR #105, opened 2016-10-31
Do not automatically throw in `continuePrimaryKey` for key == cursor position

brettz9:
Builds on prior PR related to `continuePrimaryKey`.

For `continuePrimaryKey`, do not automatically throw as with `continue`  when the key is equal to the cursor's position (the following steps will instead check that possibility, depending on the `primaryKey` and object store position)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3940,22 +3940,22 @@ The <dfn method for=IDBCursor>continuePrimaryKey(|key|,
 12. Let |primaryKey| be |r|.
 
 13. If |key| is [=less than=] or [=equal to=] this cursor's
-    [=position=] and this cursor's [=direction=] is {{"next"}}
-    or {{"nextunique"}}, [=throw=] a {{DataError}}.
+    [=position=] and this cursor's [=direction=] is {{"next"}},
+    [=throw=] a {{DataError}}.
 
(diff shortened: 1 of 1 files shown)


brettz9:
Oh, and one more thing I'm not sure how you might want fixed... In the bolded parts here:

> 15. If key is equal to this cursor’s position and primaryKey is **less than** or **equal to** this cursor’s object store position and this cursor’s direction is "next" or "nextunique", throw a DataError.
> 16. If key is equal to this cursor’s position and primaryKey is **greater than** or **equal to** this cursor’s object store position and this cursor’s direction is "prev" or "prevunique", throw a DataError.

...these link to comparisons based on *key* whereas I think they should be comparing to value.

brettz9:

inexorabletash:
Applied similar change in fdd02cf8c61025c9c79b0e193db623d12ccb3169

Thanks!

inexorabletash:
Oh, missed the "one more thing..."

The phrases you bold are the algorithms for comparing keys in general. The inputs are the _primaryKey_ argument and the cursor's _object store position_ which are both keys.

brettz9:
Thanks for your response--I've since added issue #106 so will reply there.

PR #111, opened 2016-11-04
Clear object store handle's index set when deleted. Resolves #109

### LINKED ISSUE #109: Behavior tested by transaction-abort-multiple-metadata-revert.html is not obviously specified anywhere
Behavior tested by transaction-abort-multiple-metadata-revert.html is not obviously specified anywhere

That's vague and is a mouthful, so let me explain.

This test creates some object stores, and each objectstore has an index.

At one point in the test, it:
1 - Grabs an object store handle.
2 - Deletes the references object store (which has an index)
3 - Verifies that the object store handle's set of index names is empty:
`
        database.deleteObjectStore('not_books');
...
        assert_array_equals(
            store.indexNames, [],
            'IDBObjectStore.indexNames for the deleted store3 should be empty ' +
            'immediately after IDBDatabase.deleteObjectStore() returns');
`
The problem is... I can't find where this behavior is actually specified.

Looking at the steps for IDBDatabase.deleteObjectStore, looking at behavior for "object store handle" or "index set" and other related terms... I can find nothing that states that once an object store is deleted that the object store handles referencing it should have their indexNames set cleared out.

Note: I'm not arguing with the behavior. The behavior might make sense. I see that it could go either way but don't feel strongly about it.

I'm just arguing that the behavior is either:
1 - Not specified
2 - Specified, but in a non-obvious way, and that we should make it more obvious.

inexorabletash:
Match implementation (Chrome, Firefox) and web-platform-test behavior that an object store handle's index set is cleared when the object store is deleted.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2224,7 +2224,11 @@ method, when invoked, must run these steps:
 5. Remove |store| from this [=/connection=]'s [=object
     store set=].
 
-6. Destroy |store|.
+6. If there is an [=/object store handle=] associated with
+    |store| and |transaction|, remove all entries from its
+    [=object-store-handle/index set=].
+
+7. Destroy |store|.
 
 </div>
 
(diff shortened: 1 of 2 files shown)


inexorabletash:
@beidson - does this appear correct and sufficient?

(I left "Destroy" alone for now)

beidson:
Always hard to tell in patch form, but it seems clear.

PR #117, opened 2016-11-17
Add [SameObject] annotation to global indexedDB

inexorabletash:
@foolip - can you review?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1924,7 +1924,7 @@ Indexed DB operations.
 
 <pre class=idl>
 partial interface WindowOrWorkerGlobalScope {
-  readonly attribute IDBFactory indexedDB;
+  [SameObject] readonly attribute IDBFactory indexedDB;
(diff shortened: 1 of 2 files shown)


foolip:
Is there a process for updating https://github.com/w3c/web-platform-tests/blob/master/IndexedDB/interfaces.idl or is it updated manually?

inexorabletash:
> Is there a process for updating https://github.com/w3c/web-platform-tests/blob/master/IndexedDB/interfaces.idl or is it updated manually?

Manual. I think I got started on an update a while back to try and eliminate some of the failures seen in https://bugs.chromium.org/p/chromium/issues/detail?id=435727 but then got distracted by other shiny things. Should give it another shot.

inexorabletash:
Fixed in 0f3d1fd3b3504b595ddf3871adad71151f62aad2

PR #121, opened 2016-12-02
Add non-normative documentation for methods/properties.

inexorabletash:
@aliams @domenic @bevis-tseng @beidson - what do you think about adding docs of this sort to the spec, as other specs (e.g. HTML, DOM, etc) are starting to do?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1834,7 +1834,7 @@ enum IDBRequestReadyState {
     <dt><var>request</var> . {{IDBRequest/result}}</dt>
     <dd>
         When a request is completed, returns the [=request/result=],
-        or undefined if the request failed. Throws
+        or <code>undefined</code> if the request failed. Throws
         {{InvalidStateError}} if the request is still pending.
     </dd>
 
@@ -1855,7 +1855,7 @@ enum IDBRequestReadyState {
     <dt><var>request</var> . {{IDBRequest/transaction}}</dt>
     <dd>
         Returns the {{IDBTransaction}} the request was made within.
-        If this as an [=open request=] then it returns an
+        If this as an [=open request=], then it returns an
         [=upgrade transaction=] while it is running, or null otherwise.
     </dd>
 
@@ -2679,7 +2679,7 @@ and false otherwise.
     </dt>
     <dd>
        Adds or updates a [=record=] in |store| with the given |value|
-       and |key|. If the store uses [=in-line keys=] then |key| must
+       and |key|. If the store uses [=in-line keys=], then |key| must
        not be specified, or a {{DataError}} will be thrown.
 
        If {{IDBObjectStore/put()}} is used, any existing [=record=]
@@ -2701,17 +2701,17 @@ and false otherwise.
        [=/key=] or in the given [=key range=] in |query|.
 
        If successful, |request|'s {{IDBRequest/result}} will
-       be undefined.
+       be <code>undefined</code>.
     </dd>
 
     <dt>
       <var>request</var> = <var>store</var> . {{IDBObjectStore/clear()|clear}}()
     </dt>
     <dd>
-       Deletes all records [=records=] in |store|.
+       Deletes all [=records=] in |store|.
 
        If successful, |request|'s {{IDBRequest/result}} will
-       be undefined.
+       be <code>undefined</code>.
     </dd>
   </dl>
 </div>
@@ -2924,8 +2924,9 @@ must run these steps:
        Retrieves the [=/value=] of the first [=record=] matching the
        given [=/key=] or [=key range=] in |query|.
 
-       If successful, |request|'s {{IDBRequest/result}} will
-       be the [=/value=], or undefined if there was no matching [=record=].
+       If successful, |request|'s {{IDBRequest/result}} will be the
+       [=/value=], or <code>undefined</code> if there was no matching
+       [=record=].
     </dd>
 
     <dt>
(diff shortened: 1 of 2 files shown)


aliams:
I think this is very useful and definitely helps with readability! This is especially true for those that want to get the quick gist of the expected functionality rather than having to read through lines and lines of algorithms.

inexorabletash:
@aliams - Thanks! Would you be able to take a look at the additions and provide any feedback on the text or style? The docs are not intending to cover every case (so they don't e.g. enumerate every exception) but any feedback on what's there (or what's missing!) is welcome.

(I moved one method definition so that it was grouped more logically; otherwise it's all just "domintro" additions)

aliams:
I made a quick pass-through earlier and it looked really good. I'll take another crack at it with a closer look.

index.bs
```diff
@@ -2462,6 +2663,59 @@ getter must return true if this [=/object store handle=]'s
...
+
+       If successful, |request|'s {{IDBRequest/result}} will
+       be undefined.
+    </dd>
+
+    <dt>
+      <var>request</var> = <var>store</var> . {{IDBObjectStore/clear()|clear}}()
+    </dt>
+    <dd>
+       Deletes all records [=records=] in |store|.
```
domenic: extra "records"

index.bs
```diff
@@ -2462,6 +2663,59 @@ getter must return true if this [=/object store handle=]'s
...
+    <dt>
+      <var>request</var> = <var>store</var> .
+          {{IDBObjectStore/delete()|delete}}(<var>query</var>)
+    </dt>
+    <dd>
+       Deletes [=records=] in |store| with the given
+       [=/key=] or in the given [=key range=] in |query|.
+
+       If successful, |request|'s {{IDBRequest/result}} will
+       be undefined.
```
domenic: I wonder if, in developer-facing documentation at least, `undefined` should be in code font? I'm not sure, since we have a pretty strong precedent of leaving it in normal font inside algorithms.

index.bs
```diff
@@ -2462,6 +2663,59 @@ getter must return true if this [=/object store handle=]'s
...
+      <var>request</var> = <var>store</var>
+          . {{IDBObjectStore/put()|put}}(<var>value</var> [, <var>key</var>])
+    </dt>
+    <dt>
+      <var>request</var> = <var>store</var>
+          . {{IDBObjectStore/add()|add}}(<var>value</var> [, <var>key</var>])
+    </dt>
+    <dd>
+       Adds or updates a [=record=] in |store| with the given |value|
+       and |key|. If the store uses [=in-line keys=] then |key| must
```
domenic: Comma after in-line keys would make this flow better IMO

index.bs
```diff
@@ -3177,6 +3572,37 @@ interface IDBIndex {
...
+    </dd>
+    <dt><var>index</var> . keyPath</dt>
+    <dd>
+      Returns the [=/key path=] of the index.
+    </dd>
+    <dt><var>index</var> . multiEntry</dt>
+    <dd>
+      Returns true if the index's [=index/multiEntry flag=] is set.
+    </dd>
+    <dt><var>index</var> . multiEntry</dt>
```
domenic: Copypasta

index.bs
```diff
@@ -3574,6 +4103,28 @@ interface IDBKeyRange {
...
+      Returns [=lower bound=], or undefined if none.
+    </dd>
+    <dt><var>range</var> . {{IDBKeyRange/upper}}</dt>
+    <dd>
+      Returns [=upper bound=], or undefined if none.
+    </dd>
+
+    <dt><var>range</var> . {{IDBKeyRange/lowerOpen}}</dt>
+    <dd>
+      Returns true if [=lower open flag=] is set, and false otherwise.
```
domenic: Missing "the" here and below.

index.bs
```diff
@@ -3722,6 +4314,31 @@ enum IDBCursorDirection {
...
+
+<div class=note>
+  <dl class=domintro>
+    <dt><var>cursor</var> . {{IDBCursor/source}}</dt>
+    <dd>
+      Returns the {{IDBObjectStore}} or {{IDBIndex}} the cursor was opened from.
+    </dd>
+    <dt><var>range</var> . {{IDBCursor/direction}}</dt>
+    <dd>
+      Returns the [=cursor/direction=] of the cursor.
```
domenic: Maybe parenthetically duplicating the contents of the enum might be helpful here (i.e. "next", "nextunique", ...)

index.bs
```diff
@@ -4082,6 +4781,33 @@ enum IDBTransactionMode {
...
+    <dt><var>transaction</var> . {{IDBTransaction/objectStoreNames}}</dt>
+    <dd>
+      Returns a list of the names of [=/object stores=] in the
+      transaction's [=transaction/scope=]. For an [=upgrade transaction=]
+      this is all object stores in the [=database=].
+    </dd>
+    <dt><var>transaction</var> . {{IDBTransaction/mode}}</dt>
+    <dd>
+      Returns the [=transaction/mode=] the transaction was created with
+      ({{"readonly"}} or {{"readwrite"}}) or {{"versionchange"}} for
```
domenic: Comma after parenthesis for extra clarity

index.bs
inexorabletash: I agree it makes it more readable i.e. clarifies that we're not describing undefined behavior.

inexorabletash:
Thanks @domenic !

> ...saying "InvalidStateError" DOMException is clearer for readers, especially developers...

Thanks - I'll tackle that in a follow-up.

We should get some sort of process in place for that sort of style change. I guess it should be added/modified in [Infra](https://infra.spec.whatwg.org/) and spec authors can watch for changes?

domenic:
> We should get some sort of process in place for that sort of style change. I guess it should be added/modified in Infra and spec authors can watch for changes?

I think in this case we need to change the advice in Web IDL. I'm not sure how to best announce such things, though, and agree it would be nice to get something shared in place.

inexorabletash:
@aliams @domenic - is this good enough to merge in?

domenic:
LGTM

aliams:
LGTM as well.

PR #129, opened 2016-12-16
Clarify the conditions for running create/deleteObjectStore()

inexorabletash:
The phrasing "currently running upgrade transaction" is imprecise
because "running" is not well defined. Instead, reference the
transaction lifecycle state.


inexorabletash:
@pwnall - can you review?

pwnall:
LGTM. We (Blink) will be contributing WPT tests that match this text.

PR #133, opened 2017-01-05
Adding examples for “versionchange” and “blocked”.

jakearchibald:


index.bs
```diff
@@ -278,6 +278,63 @@ request.onsuccess = function() {
 };
 </pre>
 
+The database cannot upgrade while other tabs are still using it. To avoid blocking
+other tabs, you should use the "versionchange" event to close the connection to
```
inexorabletash: should use → may listen for

index.bs
```diff
@@ -278,6 +278,63 @@ request.onsuccess = function() {
...
+    // Now you need to close the database.
+    // If the document isn't being actively used, it may be appropriate to reload:
+    if (!document.hasFocus()) {
+      location.reload();
+      // Reloading will close the database, and also reload with the new JavaScript
+      // and database definitions.
+    } else {
+      // Alternatively you may close the database and display a message to the
+      // user. You need to ensure this isn't a disruptive user experience.
+      db.close();
```
inexorabletash: maybe also `db = null` so that other page logic can notice the connection is done.

index.bs
```diff
@@ -278,6 +278,63 @@ request.onsuccess = function() {
...
+function saveUnsavedData() {
+  // How you do this depends on your app.
+}
+
+function displayMessage() {
+  // Show some sort of alert to the user.
+}
+</pre>
+
+Your new tabs can use the "blocked" event to detect if older tabs are preventing
```
inexorabletash: Your new tabs → Pages
older tabs → tabs with older versions of the code

index.bs
```diff
@@ -278,6 +278,63 @@ request.onsuccess = function() {
...
+request.onsuccess = function() {
+  hideMessage();
+};
+
+function hideMessage() {
+  // Hide a previously displayed message.
+}
+</pre>
+
+The user will only see the above message if another tab fails to disconnect from
```
inexorabletash: Maybe clarify (based on previous questions that have been raised in this area)

If all recipients of "versionchange" close the connection synchronously within the handler then "blocked" will not be fired. But if they don't close synchronously (i.e. they ignored it, or they do async work like starting a final transaction to save state) then the "blocked" event will be seen.

Once all connections finally close then "success" will be seen - that is, the blocked state is not permanent. That's why it's important to hide the message when "success" comes in.

index.bs
```diff
@@ -278,6 +278,63 @@ request.onsuccess = function() {
...
+request.onupgradeneeded = function(event) {
+  // ...
+};
+
+request.onblocked = function() {
+  displayMessage("Please close other tabs to this site.");
+};
+
+request.onsuccess = function() {
+  hideMessage();
```
inexorabletash: Should probably have `// ...` here indicating proceeding with the usual open steps

index.bs
inexorabletash: other tabs → other tabs trying to upgrade the database

index.bs
```diff
@@ -278,6 +278,63 @@ request.onsuccess = function() {
...
+  saveUnsavedData().then(function() {
+    // Now you need to close the database.
+    // If the document isn't being actively used, it may be appropriate to reload:
+    if (!document.hasFocus()) {
+      location.reload();
+      // Reloading will close the database, and also reload with the new JavaScript
+      // and database definitions.
+    } else {
+      // Alternatively you may close the database and display a message to the
+      // user. You need to ensure this isn't a disruptive user experience.
```
inexorabletash: Do we want to list the other alternative, which is to notify the user that another tab wants to upgrade, so the user should finish their work and close (or reload) this one?

(That's probably getting too long.)

marcoscaceres: I think that would be good, to be honest. I guess you could use `new BroadcastChannel("idb-messages")` or something?

marcoscaceres:
this is super helpful! Thanks @jakearchibald!

index.bs
inexorabletash: No need for BroadcastChannel - the receiver of "versionchange" could pop up a modeless notification saying e.g. _"There's a new version of this web app available! Just click reload whenever you're ready."_

marcoscaceres: ah - ok. I gotta get my head out of service worker land and message passing. 

I'm still not totally clear where .onblocked comes into play... if at all?

inexorabletash: In reality, you want something more subtle and user friendly like Jake has, such as trying to transparently reload if possible and if not then pestering the user the minimal amount.

index.bs
jakearchibald: I'll make a reference to the onclose event.

jakearchibald:
@inexorabletash excellent feedback. I've made some major changes to the original PR. The example is now in its own section with its own ID. I've covered more ways to handle "versionchange", and gone into more detail with "blocked".

index.bs
```diff
@@ -277,7 +277,94 @@ request.onsuccess = function() {
...
+  // How you do this depends on your app.
+}
+
+function displayMessage() {
+  // Show a non-modal message to the user.
+}
+</pre>
+
+Another way is to call the database's close method. Other code on your page
+should react to the database's "close" event, so it knows the database is no
```
inexorabletash: The `close` event only fires when the connection closes abnormally, i.e. when the user deletes browsing data or there is corruption or an I/O error. If `close()` is called explicitly the event does not fire.

inexorabletash: And therefore it might be worth adding a note to the `onclose` definition that it's only fired in abnormal close scenarios. Look for the _forced close_ flag.

jakearchibald:
@inexorabletash updated!

jakearchibald:
Cheers!

inexorabletash:
Thanks @jakearchibald !

I made some subsequent tweaks (wordsmithing, spelling fix, definition links) - can you take a peek at the live spec and make sure I didn't make it worse?

jakearchibald:
@inexorabletash looks good to me, thanks!

PR #134, opened 2017-01-05
Rework end-of-upgrade-transaction steps

inexorabletash:
The "steps for running an upgrade transaction" pushed behavior into the "steps for comitting/aborting"  committing transactions. Make those explicit steps in the latter algorithms instead.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5517,7 +5517,7 @@ for the [=database=], and a |request|.
 8. Wait for |transaction| to [=transaction/finish=].
 
     <aside class=note>
-      Some of algorithms invoked during the [=/transaction=]'s
+      Some of the algorithms invoked during the [=/transaction=]'s
       [=transaction/lifetime=], such as the [=steps for committing a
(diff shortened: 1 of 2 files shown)


inexorabletash:
@smaug---- - could you take a peek?

index.bs
```diff
@@ -5405,13 +5416,12 @@ for the [=database=], and a |request|.
...
 
-9. If |transaction| is aborted for any reason, the [=steps for
-    aborting a transaction=] must be run.
-
-10. When |transaction| is finished, immediately set |request|'s
-    [=request/transaction=] to null. This must be done in the same
-    task as the task firing the <code>complete</code> or
-    <code>abort</code> event, but after the event has been fired.
+    <aside class=note>
+      Some of algorithms invoked during the [=/transaction=]'s
```
aliams: Some of _the_ algorithms...

aliams:
I agree that this makes it more explicit.

PR #136, opened 2017-01-13
Remove DOMStringList definition

inexorabletash:
Following https://github.com/whatwg/html/commit/a830aef7a9087b0d879f548980f0c616df02149e the definition of DOMStringList is now in [[HTML]](https://html.spec.whatwg.org/multipage/infrastructure.html#the-domstringlist-interface) - remove it from here.


inexorabletash:
@zcorpan - can you sanity check?

Issue #135, opened 2017-01-10
Update usage of StructuredClone


inexorabletash:
The spec abstractly references the _structured clone_ algorithm, e.g. _Let clone be structured clone of value..._ - this should be updated to use the **StructuredClone(_input_,_targetRealm_)** syntax from [[HTML]](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredclone).

* [put()](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-put), [add()](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-add) and [update()](https://w3c.github.io/IndexedDB/#dom-idbcursor-update) should call StructuredClone() explicitly using the current Realm
* [Object Store Retrieval Operations](https://w3c.github.io/IndexedDB/#object-store-retrieval-operation), [Index Retrieval Operations](https://w3c.github.io/IndexedDB/#index-retrieval-operation) and [Cursor Iteration Operation](https://w3c.github.io/IndexedDB/#index-retrieval-operation) should take a _targetRealm_ and use it when calling StructuredClone(); callers should pass the current Realm

PR #137, opened 2017-01-18
Use StructuredClone() hook

inexorabletash:
This is for https://github.com/w3c/IndexedDB/issues/135 - switch from prose to algorithmic invocations of [StructuredClone](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredclone)

@domenic - could you review this?

I'm not sure why the a manual dfn of StructuredClone is necessary.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -36,7 +36,6 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: worker; url: worker
     urlPrefix: infrastructure.html
         type: dfn
-            text: structuredclone; url: structuredclone
             text: strictly splitting; url: strictly-split-a-string
 spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
     type: dfn
@@ -785,7 +784,7 @@ from a [=/value=]. A <dfn>valid key path</dfn> is one of:
 </aside>
 
 [=/Key path=] values can only be accessed from properties explicitly
-copied by the [=StructuredClone=], as well as the
+copied by <a abstract-op>StructuredClone</a>, as well as the
 following type-specific properties:
 
 <table class=props>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -36,9 +36,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: worker; url: worker
     urlPrefix: infrastructure.html
         type: dfn
+            text: structuredclone; url: structuredclone
```
domenic: You should be able to not do this if you use `<a abstract-op>StructuredClone</a>(...)` instead of `[=StructuredClone=](...)`. This will also ensure it shows up in your indices with the correct casing.

index.bs
```diff
@@ -785,7 +785,7 @@ from a [=/value=]. A <dfn>valid key path</dfn> is one of:
 </aside>
 
 [=/Key path=] values can only be accessed from properties explicitly
-copied by the [=structured clone algorithm=], as well as the
+copied by the [=StructuredClone=], as well as the
```
domenic: remove "the"

inexorabletash:
> I assume the "user agent defined realm" situations are for when you're serializing to disk or similar?

Exactly. The model is:

If you're passing data to Indexed DB it does an immediate clone (which may have side effects, fail, throw, etc); the result may be used for various operations (e.g. extracting index keys) but that must be unobservable and side-effect free, and eventually a serialized form hits disk. So I lifted "user-agent defined Realm" from postMessage(), which seems to capture that well.

If you're retrieving data from Indexed DB, it's within the context of the request so the current Realm gets passed along as targetRealm.

index.bs
inexorabletash: Thanks!

PR #143, opened 2017-02-06
Fix markdown indentation to repair algorithm. Resolves #142

### LINKED ISSUE #142: Step for multientry key conversion are mis-indented
https://w3c.github.io/IndexedDB/#convert-a-value-to-a-multientry-key

> Return a new array key with value set to a list of the members of keys.

This should be part of the "If IsArray(input), then:" substeps.

inexorabletash:
Literally just whitespace causing an algorithm to have a step in the wrong place.

Also added the "run these substeps" boilerplate, and fixed a couple ECMAScript spec refs.


inexorabletash:
@aliams - can you review?

PR #145, opened 2017-02-13
Implement transaction cleanup hook for [HTML]

inexorabletash:
The other side of https://github.com/whatwg/html/pull/2310

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -966,6 +966,9 @@ if new [=requests=] can be made against the transaction. A
 transaction is said to be <dfn>active</dfn> if its [=transaction/active flag=]
 is set.
 
+A [=/transaction=] optionally has a <dfn>cleanup event loop</dfn>
+which is an [=event loop=].
+
 A [=/transaction=] has a <dfn>request list</dfn> of [=requests=]
 which have been made against the transaction.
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1143,6 +1146,25 @@ The following constraints define when a [=/transaction=] can be
 
 </div>
 
+To <dfn export>cleanup transactions</dfn> for an [=event loop=], run
```
annevk: Could you call this "cleanup Indexed Database transactions"? Or do you want HTML to call it "cleanup transactions"? I don't think we should use different terms.

inexorabletash: Can do.

index.bs
```diff
@@ -1143,6 +1146,25 @@ The following constraints define when a [=/transaction=] can be
 
 </div>
 
+To <dfn export>cleanup Indexed Database transactions</dfn> for an
+[=event loop=], run the following steps for each [=/transaction=]
+with a matching [=transaction/cleanup event loop=]:
```
annevk: You need to grab the event loop from the current settings object, right @domenic? We don't pass it in.

domenic: All specs I've seen have just used a global notion of event loop. Which seems reasonable; there's only one per process.

inexorabletash: Can you suggest text here? Just use "the event loop", or "the relevant event loop" or ...?

domenic: IMO "the [=event loop=]" is fine.

index.bs
```diff
@@ -1143,6 +1146,25 @@ The following constraints define when a [=/transaction=] can be
 
 </div>
 
+To <dfn export>cleanup Indexed Database transactions</dfn> for the
+[=event loop=], run the following steps for each [=/transaction=]
+with a matching [=transaction/cleanup event loop=]:
```
inexorabletash: I'm still not thrilled with this. Pondering something like: "To __cleanup Indexed Database transactions__, run the following steps for each __transaction__ with __cleanup event loop__ matching the __event loop__:"

That's still awkward phrasing, though. 

(The spec needs to speak about transactions that span multiple threads, although each transaction is of course associated with just one. In most cases the spec is written either in the context of one thread, or in the context of the "database thread" and so dodges the complexity. It's never explicit about this, though.)

domenic: Hmm, I guess in this case I'd be OK saying your version but with "... matching the current event loop"? It's not really a phrase used elsewhere but it helps make things easier to read here so it works for me.

inexorabletash:
Tweaked the text a bit more to use "current event loop".

Still haven't landed tests yet.

inexorabletash:
Tests are in at https://github.com/w3c/web-platform-tests/commit/57aa2ac737eec9526ad6c4ace61e590730ec3b9e

Issue #144, opened 2017-02-12
Simplify "steps for storing a record into an object store"?


brettz9:
Unless I'm missing a conceptual or anticipated reason for specifying the steps this way, I'd think that step 1 of the "steps for storing a record into an object store" routine (i.e., "If store uses in-line keys, run these substeps:") could be removed and thereby limit some redundancy by having `IDBCursor.update` set `key` to `kpk` in its "If the effective object store of this cursor uses in-line keys, run these substeps:" routine immediately preceding its call to the record storing steps (as do `IDBObjectStore.put` and `IDBObjectStore.add` in their equivalent, corresponding step preceding the record storing steps), and then supply that `key` instead of the effective key as `key` to the record storing steps in such a case (assuming it is not a failure).

I know implementations could optimize the redundancy out anyways, but it seems it would be simpler conceptually as well unless I'm missing a reason for it.

brettz9:
Btw, one nit for step 5...

> such that the list is sorted according key of the records in ascending order

should presumably be:

> such that the list is sorted according to the key of the records in ascending order

brettz9:
And while I'm on the subject, for one other minor problem I see in the record storing steps, I think

> Otherwise, if the type of key is number and the value is greater than or equal to the key generator's current number, set the current number to lowest integer greater than key.

should be:

> Otherwise, if the type of key is number and the value is finite and greater than or equal to the key generator's current number, set the current number to the lowest integer greater than key.

brettz9:
And if an exception is thrown in `inject a key into a value using a key path`, the consuming record storing steps don't indicate what to do with that exception.

brettz9:
And I believe:

> Let keyValue be the result of running the steps to convert a key to a value.

ought to be:

> Let keyValue be the result of running the steps to convert a key to a value with `key` as `key`.

inexorabletash:
Thank you very much @brettz9 for finding these. Nothing like a fresh implementation to shake out the glitches. I'll try and get a PR up this week to address these.

inexorabletash:
>  ...having IDBCursor.update set _key_ to _kpk_...

This is not needed, as that block is responsible just checking that the extracted _key_ equals _effective key_. If not equal, the steps abort since `update()` is not allowed to modify a key. And_effective key_ is already passed as _key_. 

But agreed, step 1 in "steps for storing a record into an object store" is redundant.

----

> ... according to the key ...

Yep, will fix.

----

> ... value is finite ...

[...]

PR #146, opened 2017-02-13
Fixes for 'steps for storing'

inexorabletash:
@brettz9 - could you do an initial review?

This covers the issues from #144 (apart from the > MAX_SAFE_INTEGER/keygen ones)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2941,8 +2941,15 @@ when invoked, must run these steps:
 
     3. If |kpk| is not failure, let |key| be |kpk|.
 
-    4. Otherwise, if |store| does not have a [=key
-        generator=], [=throw=] a "{{DataError}}" {{DOMException}}.
+    4. Otherwise (|kpk| is failure), run these substeps:
+
+        1. If |store| does not have a [=key generator=], [=throw=]
+            a "{{DataError}}" {{DOMException}}.
+
+        2. Otherwise, if the steps to
+            [=check that a key could be injected into a value=] with
+            |clone| and  |store|'s [=object-store/key path=] return
+            false, [=throw=] a "{{DataError}}" {{DOMException}}.
 
 12. Run the steps to [=asynchronously execute a request=] and
     return the {{IDBRequest}} created by these steps. The steps are
@@ -3006,8 +3013,15 @@ when invoked, must run these steps:
     3. If |kpk| is not failure, let |key| be
         |kpk|.
 
-    4. Otherwise, if |store| does not have a [=key
-        generator=], [=throw=] a "{{DataError}}" {{DOMException}}.
+    4. Otherwise (|kpk| is failure), run these substeps:
+
+        1. If |store| does not have a [=key generator=], [=throw=]
+            a "{{DataError}}" {{DOMException}}.
+
+        2. Otherwise, if the steps to
+            [=check that a key could be injected into a value=] with
+            |clone| and  |store|'s [=object-store/key path=] return
+            false, [=throw=] a "{{DataError}}" {{DOMException}}.
 
 12. Run the steps to [=asynchronously execute a request=] and
     return the {{IDBRequest}} created by these steps. The steps are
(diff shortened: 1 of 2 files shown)


brettz9:
It all looks go to me, with the exception that I'm unclear on the following...

It seems to me that `3. If |hop| is false, return true.` should instead be:

```
>3. If |hop| is false, run these substeps
>    1. Let o be a new Object created as if by the expression ({}).
>    2. Let status be CreateDataProperty(value, identifier, o).
>    3. Assert: status is true.
```

...because the whole algorithm ought to continue on to keep checking the remaining identifiers. However, I guess you will need to add a step to make a clone of this clone if the checking could have inadvertent side effects.

inexorabletash:
> ...because the whole algorithm ought to continue on to keep checking the remaining identifiers.

If _hop_ is false it has no property; the corresponding injection steps will add a blank object as necessary for the remaining (non-last) identifiers. That blank object will have no own-properties so therefore it will (inductively) be okay to add more.

brettz9:
I see, then LGTM.

Btw, one bit of personal info I might share in case it may be useful in future collaborations... I suffer from CFS with attendant brain fog, so although I have the underlying ability to think these kinds of things through, I can get a kind of tunnel vision (beyond that we all experience). I try to compensate by testing and revisiting as well as generally trying hard to think something through before posting, but it may sometimes come out seeming like I'm just being careless or lazy. I just mention it in case it may minimize any frustration in communications (and hopefully any other contributions I can make will compensate).

inexorabletash:
@brettz9 - having a diversity of skills, interests, abilities, backgrounds, and communication and working styles is incredibly valuable to producing the best possible specs. You've noticed several issues that others missed. Your contributions are very much appreciated!

inexorabletash:
@aliams - can you review these changes?

brettz9:
Realized I have a few more questions that may be somewhat interconnected if you want them here or before completing the PR:

1. Should the second step of "convert a value to a multiEntry key", namely:

> Otherwise, return the result of running the steps to convert a value to a key with argument input.

become:

> Otherwise, return the result of running the steps to convert a value to a key with argument input. Rethrow any exceptions.

...or is the rethrowing somehow assumed with "return"?

2. For "extract a key from a value using a key path", should the second step, namely:

> If r is failure, return failure.

be followed by:

> If r is invalid, return invalid.

(or an assertion if this is not possible as one might think could occur given that "evaluate a key path on a value" recurses back to these steps which can return "invalid")?

3. As with #2, should "evaluate a key path on a value", should step 1.3.1, namely:

> Let key be the result of recursively running the steps to extract a key from a value using a key path using item as keyPath and value as value.

...be followed by a condition about what to do if it is valid (or assertion that it should not be possible as one might expect given that "extract a key from a value using a key path" can return "invalid")

brettz9:
And though for a different algorithm, one other little spec issue:

> Let octets be the result of running the steps to get a copy of the bytes held by the buffer source `value`.

ought to be:

> Let octets be the result of running the steps to get a copy of the bytes held by the buffer source `input`.

inexorabletash:
> 1. Should the second step of "convert a value to a multiEntry key"....

Fixed.

> 2. For "extract a key from a value using a key path" ...

> 3. As with #2, should "evaluate a key path on a value", should step 1.3.1...

That should recurse to "evaluate" directly, not "extract". Fixed.

> held by the buffer source input.

Fixed.

brettz9:
Excellent, thanks! LGTM...

index.bs
```diff
@@ -6419,8 +6465,7 @@ as follows. The algorithm takes a |value|, a |key| and a |keyPath|.
 4. For each remaining |identifier| in |identifiers|, run these
     substeps:
 
-    1. If |value| is not an [=Object=] object or an [=Array=]
-        object, then [=throw=] a "{{DataError}}" {{DOMException}}.
+    1. Assert: |value| is an [=Object=] or an [=Array=].
```
aliams: I remember we explicitly wanted to throw a DOMException in this case. Does asserting give us the same result here?

brettz9: See the comment below that 

> Assertions can be made in the above steps because this algorithm is
> only applied to values that are the output of <a abstract-op>StructuredClone</a>,
> and the steps to [=check that a key could be injected into a value=] have
> been run.

The latter steps to "check that a key could be injected into a value" were added (prior to these steps) so that throwing would occur synchronously as opposed to within these record storing steps you have highlighted where they would be occurring asynchronously with the request (and where throwing would not make sense).

inexorabletash: Right. "Assert" is just a guide to implementers to help with the readability of algorithms. Throwing required breaking out the steps separately, as @brettz9 clarifies.

aliams: Thank you @brettz9 and @inexorabletash for clarifying!

Issue #148, opened 2017-02-14
specify what should happen if the origin of execution context is opaque


jeisinger:
right now, there is no explicit definition of what should happen, however, both blink and webkit will through a security error during IDBFactory::open if the origin is opaque.

inexorabletash:
See also https://github.com/whatwg/storage/issues/41 which I filed this morning.

PR #150, opened 2017-02-14
open/deleteDatabase should throw on opaque origin

inexorabletash:
Context: https://github.com/w3c/IndexedDB/issues/148

Needs tests.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2120,13 +2120,18 @@ when invoked, must run these steps:
 
 1. If |version| is 0 (zero), [=throw=] a [=TypeError=].
 
-2. Let |request| be a new [=open request=].
+2. Let |origin| be the [=/origin=] of the global scope used
+    to access this {{IDBFactory}}.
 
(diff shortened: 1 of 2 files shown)


inexorabletash:
The text _"the origin of the global scope used to access this IDBFactory"_ could almost certainly be improved, but it's what was there already.

inexorabletash:
@aliams - could you review?

foolip:
Tests for this are at https://github.com/w3c/web-platform-tests/pull/4832, it would be ideal if the spec change is merged before the tests, but that's blocking Chromium's test export process so I'll probably go ahead and merge that first. It will likely be reverted though if the test change isn't merged soon after.

inexorabletash:
I guess we merge spec changes before landing test changes?

@foolip - can you review https://github.com/w3c/web-platform-tests/pull/4829 ?

foolip:
OK, so there were other tests as well. I'll review those.

PR #151, opened 2017-02-15
Editorial: remove repeated word

pra85:
Removing the repeated word - `using`


pra85:

inexorabletash:
Thanks! I'll regenerate the HTML soon.

Issue #147, opened 2017-02-13
Key generator behavior with explicit keys above 2^53


inexorabletash:
cc: @bevis-tseng @aliams @beidson

@brettz9 noticed this in #144 as a spec oversight, but implementations are inconsistent and the desired behavior is not obvious to me, so calling this out separately. 

Test case is included in https://github.com/w3c/web-platform-tests/pull/4814

As background, if you have a key generator (stored created with `autoIncrement: true`) you can explicitly advance the generator, e.g.
```js
store.put(value); // gets generated key, e.g. 1
store.put(value, 1000); // gets explicit key and sets generator to 1000
store.put(value); // gets generated key, i.e. 1001
store.put(value, 555); // gets explicit key but is < generator so no change to generator
store.put(value); // gets generated key, i.e. 1002
```
So far so good.

Now what happens in the following cases:

* Explicit key is greater than the max generator value (2^53 i.e. Number.MAX_SAFE_INTEGER)
* Explicit key is at 2^63 (max int64_t)
* Explicit key is at 2^64 (max uint64_t)
* Explicit key is greater than 2^64
* Explicit key is Infinity

e.g. for Infinity:
```js
store.put(value, Infinity); // gets explicit key and sets generator to ????
store.put(value).onsuccess = e => console.log(e.target.result); // ????
store.put(value).onsuccess = e => console.log(e.target.result); // ????
```

It looks like implementations are inconsistent and outright buggy here. With Infinity:

* Firefox generates -9223372036854776000 as the next key (!), then ConstraintError for subsequent puts (!!)
* Safari resets the generator to 0, so you start getting 1, 2, 3, etc. again (!)
* Chrome does not modify the generator, but that's only "by accident".

For reference, Chrome internally coerces the input Number (double) to an int64_t before checking to see if it should increment the generator (also an int64_t) so our behavior goes wonky above the 53 bit mark but at the 63 bit mark it ends up being ignored. Wheee.

There are two possible behaviors for any of these ranges of values:

1. Update generator to maximum value (same as if maximum value was passed)
2. Do not modify the generator (same as if a anything lower than current number was passed)

I definitely think treating bleeding implementation details and treating different finite values above the generator range inconsistently is bad. (i.e. I want to fix Chrome here). I don't particularly care what we do for finite vs. infinite values - they could both be (1), or (2), or differ. 

Thoughts?

aliams:
I think updating to the maximum value makes the most sense here.

[...]

PR #153, opened 2017-02-22
Rework key generator algorithm

inexorabletash:
Address the spec gaps pointed out in https://github.com/w3c/IndexedDB/issues/147

This pins the key generator to the maximum value (2^53). The logic around incrementing/updating the generator is made algorithmic rather than being duplicated between the object store storage operation steps and the prose definition of the key generator.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1494,88 +1494,130 @@ stated otherwise it is unset.
 <!-- ============================================================ -->
 
 When a [=/object store=] is created it can be specified to use a
-<dfn>key generator</dfn>. A [=key generator=] keeps an internal
-<dfn>current number</dfn>. The [=current number=] is always a positive
-integer. Whenever the key generator is used to generate a new
-[=/key=], the generator's [=current number=] is returned and
-<strong>then</strong> incremented to prepare for the next time a new
-[=/key=] is needed. Implementations must use the following rules for
-generating numbers when a [=key generator=] is used.
+<dfn>key generator</dfn>. A key generator is used to generate keys
+for records inserted into an object store if not otherwise specified.
 
-* Every object store that uses key generators use a separate
-    generator. I.e. interacting with one object store never affects
-    the key generator of any other object store.
+<div dfn-for="key generator">
 
-* The [=current number=] of a [=key generator=] is always set to
-    1 when the [=/object store=] for that key generator is first
-    created.
+A [=key generator=] has a <dfn>current number</dfn>. The
+[=key generator/current number=] is always a positive integer less
+than or equal to 2<sup>53</sup> (9007199254740992). The initial value
+of a [=key generator=]'s [=key generator/current number=] is 1, set
+when the associated [=/object store=] is created. The
+[=key generator/current number=] is incremented as keys are generated,
+and may be updated to a specific value by using explicit keys.
 
-* When a key generator is used to generate a new [=/key=] for
-    a [=/object store=], the key generator's [=current
-    number=] is used as the new key value and then the key
-    generator's [=current number=] is increased by 1.
+</div>
 
-* When a [=object-store/record=] is stored and a [=/key=] is specified
-    in the call to store the record, if [=key/type=] of the key
-    is <i>number</i> and the [=key/value=] is greater than or
-    equal to the key generator's [=current number=], then the key
-    generator's [=current number=] is set to the smallest integer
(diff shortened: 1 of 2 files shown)


inexorabletash:
@aliams can you review?

brettz9:
From my perspective, it looks good, but I have a few concerns:

1. Nit: I think in the context of spec language, "possibly update" sounds better than "maybe update".

2. I believe in the algorithm to "maybe update the key generator":

> 5. If |value| is greater than |generator|'s [=key generator/current number=],
>       then set |generator|'s [=key generator/current number=] to |value|.

should be something like the following (unless the former behavior is being changed):

> 5. If |value| is greater than |generator|'s [=key generator/current number=],
>       then set |generator|'s [=key generator/current number=] to |value| (rounded up to the nearest integer if it has a decimal value).

(Note that it is safe to use this language since you set `value` before this step to a maximum of 2^53.)

3. In the algorithm to "generate a key", with these steps:

> 3. If |key| is greater than 2<sup>53</sup> (9007199254740992), then return failure.
>
> 4. Increase |generator|'s  [=key generator/current number=] by 1.

...an implementation note should perhaps be added to mention that if `key` is already at 9007199254740992, then incrementing by one in the likes of ECMAScript will still lead to 9007199254740992 and thus failure will never be reached (and a value may be unintentionally overwritten). An alternative to adding an implementation note (and I think a safer one) would be changing the max instead to Number.MAX_SAFE_INTEGER (i.e., 2^53 - 1).

inexorabletash:
Thanks @brettz9 !

Yes, I missed both the floor() step _and_ incrementing. :( 

I've hopefully corrected the steps, and made a note about where the limit comes from, although I didn't go into "what if..." detail. 

Re: altering the maximum:

We already have tests around the boundary condition and that the generator may produce 9007199254740992 so I don't think we want to reduce that. In a pure JS implementation I scraped together to sanity check my logic, I used e.g.:
```js
    this.current += 1;
    if (this.current === key)
      this.current = Infinity;
```
and
```js
    if (value >= this.current) {
      this.current = value + 1;
      if (this.current === value)
        this.current = Infinity;
    }
```
An implementation using `double`s could do that as well. I expect most implementations use an `int64` or `uint64` internally and can use the increment beyond 9007199254740992 safely.

brettz9:
Actually, you don't want to add the incrementing step there.  Those steps you added are for *setting* the key generator when the key is undefined.

(Maybe it would be more clear to say "possibly update the current number" or "possibly update the key generator current number" instead of "possibly update the key generator" since the latter has stronger connotations of incrementing.)

The other changes fixes looked good.

inexorabletash:
>  ...I had forgotten even this step should include incrementing.

As had I. :)

brettz9:
I see one other issue...

The current number definition states the following:

> A [=key generator=] has a <dfn>current number</dfn>. The
> [=key generator/current number=] is always a positive integer less
> than or equal to 2<sup>53</sup> (9007199254740992).

However the steps to "generate a key" and "possibly update the key generator" both allow for a current number to be set to a number one higher than 9007199254740992 as does the following:

> When the [=key generator/current number=] of a key generator reaches above the
> value 2<sup>53</sup> (9007199254740992) any attempts to use the
> key generator to generate a new [=/key=] will result in a
> "{{ConstraintError}}" {{DOMException}}.

I believe the current number definition ought to be reworded, e.g., like this:

>> A [=key generator=] has a <dfn>current number</dfn>. The
> [=key generator/current number=] is always a positive integer less
> than or equal to 2<sup>53</sup> + 1 (9007199254740993).

brettz9:
Perhaps this text:


> When the [=key generator/current number=] of a key generator reaches above the
> value 2<sup>53</sup> (9007199254740992) any attempts to use the
> key generator to generate a new [=/key=] will result in a
> "{{ConstraintError}}" {{DOMException}}.

ought to be clarified as well to state:


> When the [=key generator/current number=] of a key generator reaches above the
> value 2<sup>53</sup> (9007199254740992) any subsequent attempts to use the
> key generator to generate a new [=/key=] will result in a
> "{{ConstraintError}}" {{DOMException}}.

inexorabletash:
Nice, thanks. I'll incorporate those suggestions.

inexorabletash:
Tests coming in https://github.com/w3c/web-platform-tests/pull/5065

PR #162, opened 2017-03-04
Rework event dispatching

inexorabletash:
Use 'fire an event' steps when easy, proper event creation and 'dispatch' steps otherwise.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2042,13 +2042,20 @@ Events are constructed as defined in <a
 href="https://www.w3.org/TR/domcore/#constructing-events">Constructing
 events</a>, in [[!DOM]].
 
-To <dfn>fire a version change event</dfn> named |e| at |target| given
-|oldVersion| and |newVersion|, dispatch an event at |target|. The
-event must use the {{IDBVersionChangeEvent}} interface with its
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -2155,17 +2162,13 @@ when invoked, must run these steps:
...
+            of |request| to |result| and [=fire an event=] named
+            <code>error</code> at |request| with its
+            {{Event/bubbles}} attribute initialized to true.
 
         2. Otherwise, set the [=request/result=] of |request|
-            to |result| and dispatch an event at |request|. The event
-            must use the [=Event=] interface and set the
-            {{Event/type}} attribute to "<code>success</code>". The
-            event does not bubble and is not cancelable. If the steps
+            to |result| and [=fire an event=] named
```
pwnall: Is this different from the "fire a success event" algorithm defined below?

inexorabletash: Yes, see the note just below this algorithm. _"The firing of..."_

pwnall: Thank you for clarifying! The GitHub edit window is too small... I'll review the changes with the HTML on the right.

pwnall:
LGTM.

PR #164, opened 2017-03-09
open/deleteDatabase error events are cancelable in implementations

inexorabletash:
The spec defines the error events thrown from open() and deleteDatabase() with cancelable false, but implementations have this as true. Align the spec with reality.

Also:
* Make various asides into details with summaries - for "why aren't the normal fire a success/error event steps used?", and to explain why the name and version getters are convoluted.
* Fix the autolinking for deleteIndex
* Flesh out the change history with recent big changes.

Tests: https://github.com/w3c/web-platform-tests/pull/5108


inexorabletash:
@pwnall - can you review?

Issue #167, opened 2017-03-17
Don't use sequence<>, DOMString, and DOMStringList outside of bindings


inexorabletash:
For example, the definition of [key path](https://w3c.github.io/IndexedDB/#key-path-construct) has:

> A key path is a `DOMString` or `sequence<DOMString>` that defines how to extract a key from a value.

This should use abstract spec types instead, i.e. "a string or a list of strings"

Places to update:

* [ ] [sorted list](https://w3c.github.io/IndexedDB/#sorted-list) - change to just a list which is sorted, convert to `DOMStringList` at "call site"
* [ ] [Key Path](https://w3c.github.io/IndexedDB/#key-path-construct) - "string or list of strings", etc.
* [ ] [IDBObjectStore's keyPath getter](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-keypath) - e.g. "as a `DOMString` (if a string) or `sequence<DOMString>` (if a list of strings) as appropriate"
* [ ] [IDBIndex's keyPath getter](https://w3c.github.io/IndexedDB/#dom-idbindex-keypath) - ditto
* [ ] [evaluate a key path on a value](https://w3c.github.io/IndexedDB/#evaluate-a-key-path-on-a-value) - similarly, just use "list of strings"

PR #168, opened 2017-03-17
Don't bleed IDL types into constructs/algorithms

inexorabletash:
Avoid defining constructs like 'key path' in terms of IDL types like 'DOMString' and 'sequence<>'. Instead use abstract spec types like 'string' and 'list'.

* "sorted list" is now just a list - getters mention the conversion to DOMStringList
* key path is a "string or list of strings" - getters mention the conversion to IDL types

https://github.com/w3c/IndexedDB/issues/167

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -361,7 +361,7 @@ from the database. Ideally the user will never see this.
 <h2 id=constructs>Constructs</h2>
 <!-- ============================================================ -->
 
-A <dfn>sorted list</dfn> is a {{DOMStringList}} containing strings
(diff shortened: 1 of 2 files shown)


inexorabletash:
@domenic - maybe you can review this and see if it is worth it?

(sorry that I don't have fancy preview stuff hooked up for IDB)

domenic:
Seems like a good idea to me. Feel free to link to Infra types BTW with `[=list=]` and similar.

inexorabletash:
Thanks - not linking for now but will do a pass at some point.

PR #169, opened 2017-03-19
fix code example regarding preventDefault()

jeromehode:


siusin:

inexorabletash:
Thanks! I'll regenerate the HTML.

Issue #170, opened 2017-03-20
Update for structured cloning changes in HTML


domenic:
In https://github.com/whatwg/html/commit/97d644c97335956610a31e8ad98d1a388c063e84 HTML changed the setup around structured cloning significantly, in a way that will hopefully benefit IDB a lot. You can now use the StructuredSerialize and StructuredDeserialize operations directly.

Let me know if I can help with this.

inexorabletash:
Thanks. PRs would be welcome, of course, but I'll wrangle it. :) 

The fun bit is where we currently do a clone then (1) look up properties against the clone to avoid side effects - primary key at https://w3c.github.io/IndexedDB/#dom-idbobjectstore-put and index keys at https://w3c.github.io/IndexedDB/#store-a-record-into-an-object-store which looks at the index keys, and (2) inject auto-generated keys in https://w3c.github.io/IndexedDB/#store-a-record-into-an-object-store 

I'll have to see if this can be expressed in terms of serialized records or if we will end up doing a deserialization and reserialization.


. That needs to be a serialize to _record_, deserialize _record_ to _clone_, then discard _clone_ and store _record_ on disk.

annevk:
The other thing Indexed DB should probably define is some kind of Store operation, to define how these records end up on disk (and what further changes need to be made for that).

inexorabletash:
> define how these records end up on disk

Are you thinking of hooks for Storage (which makes sense - e.g. checking for sufficient quota / updating usage), or something specific to serializing the records (which seems like an implementation detail to me)?

annevk:
In particular I think it's not quite an implementation-detail for File, Blob, and SharedArrayBuffer objects. E.g., does File continue to be a pointer or do we make a copy and the same question applies to SharedArrayBuffer. I think the likely answers are different for both.

But it would indeed also be good to integrate with Storage (which also makes it clearer what happens if you clear a box, that Indexed DB disappears too).

inexorabletash:
Of course, this had to materialize the day we were planning to [advance Indexed DB 2.0 to CR](https://github.com/w3c/IndexedDB/issues/165). *sigh*

inexorabletash:
https://github.com/w3c/IndexedDB/pull/171 for the minimum/initial set of changes, without digging into the work @annevk mentions

inexorabletash:
[...]

PR #171, opened 2017-03-22
Use StructuredSerialize/StructuredDeserialize

inexorabletash:
So far as I can tell, this is the minimum set of changes needed to replace uses of StructuredClone with StructuredSerialize/StructuredDeserialize, per #170 

Note that when `put()` (etc) is called that this does not simply serialize then operate on the output record -
the record is immediately deserialized to a _clone_ in an abstract _targetRealm_ so that subsequent operations (extract a key, inject a key, etc) are not updated. I added non-normative details explaining this at the various sites.

(Somewhat coincidentally, that's how Blink is actually implemented, since the output of serialization is a opaque set of bytes rather than an easily-traversed structure.)

The definition of operations that operate on the _clone_ could be changed to operate on the record instead. This would require redoing the "extract a key from a value" and "inject a key into a value" section (replace "value" with "record"), and duplicating the "convert a key to a value" and "convert a value to a key" steps with "record" variants.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -28,11 +28,6 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: dom.html
         type: interface
             text: Document; url: document
-    urlPrefix: infrastructure.html
-        type: abstract-op
-            text: StructuredClone; url: structuredclone
-        type: dfn
-            text: cloneable objects; url: cloneable-objects
 spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
     type: dfn
         url: sec-algorithm-conventions
@@ -61,6 +56,7 @@ spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
         text: Realm; url: realm
         text: current Realm; url: current-realm
         text: Array.prototype.sort; url: sec-array.prototype.sort
+        text: Record; url: sec-list-and-record-specification-type
 spec: webidl; urlPrefix: https://heycam.github.io/webidl/
     type: dfn
         text: sequence<DOMString>; url: idl-sequence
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -2954,9 +2964,17 @@ when invoked, must run these steps:
...
+10. Let |clone| be a [=clone=] of |value| in |targetRealm|.
     Rethrow any exceptions.
 
+    <details class=note>
+      <summary>Why create a copy of the value?</summary>
+      The value must be serialized when stored. Treating it as a copy
+      here allows other algorithms in this specification to treat it as
+      an ECMAScript value, but implementations may optimize this
+      if the difference in behavior is not observable.
+    </details>
```
annevk: This difference can be observable, especially with very large ArrayBuffer objects. With those the initial copy might succeed, but then a later serialization could fail in a naive implementation. That seems somewhat bad.

index.bs
```diff
@@ -28,6 +28,12 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: dom.html
         type: interface
             text: Document; url: document
+    urlPrefix: infrastructure.html
```
domenic: All of these should link automatically without having to be added here. Are they not?

index.bs
domenic: I do think it'd be ideal to operate on the Records in the spec, although it is interesting that Blink doesn't do that.

It would also allow you to get rid of this user-agent-defined Realm business. Part of the motivation for this overhaul was to allow specs to stop doing that sort of thing.

I can understand wanting to land a minimal version that "un-breaks" the spec first though, so working on that as a follow-up seems OK.

index.bs
```diff
@@ -5816,7 +5868,8 @@ follows.
     to=] |key|, then remove the [=object-store/record=] from |store| using the
     steps to [=delete records from an object store=].
 
-4. Store a record in |store| containing |key| as its key and |value|
+4. Store a record in |store| containing |key| as its key and
+    <a abstract-op>StructuredSerialize</a>(|value|)
```
domenic: What happens if this throws?

index.bs
inexorabletash: Not as of yesterday... looks like bikeshed has picked them up now, though.

index.bs
inexorabletash: That should get a `!` sigil. At this point there are several layers of spec fiction happening. It's "safe" because this is operating on the clone made into a user-agent-defined Realm which has only been mutated by the previous steps. 

In Blink, at least, since these steps are running in a process w/o a JS implementation, we actually do a bunch of the work (computing index keys) earlier, and some of the work (injecting generated keys) lazily when the values are later retrieved! Single-process implementations probably work more like the spec does (or have observable behavior differences/bugs), but I haven't looked at the code for any recently.

@domenic - I'll try and grab time with you to discuss some of this in real-time; I think we're hitting the limits of communicating through the peephole of review comments. :) Obviously the specs should be free of assumptions, but getting some advice on approaches for spec fiction would be swell. That wouldn't be in lieu reviews for actual changes, of course.

Issue #179, opened 2017-03-23
sorted list by code unit


aphillips:
https://www.w3.org/TR/IndexedDB-2/#constructs

> A sorted list is a DOMStringList containing strings sorted in ascending order by code unit.

While sorting by code unit is easy to do in (UTF-16) DOMString, it produces a perverse order for supplementary characters (i.e. those > U+FFFF). Shouldn't this be sorted by codepoint?

Note too that a list sorted by code unit/codepoint will not be accessible to users.

Note that there is a sorting algorithm [here](https://www.w3.org/TR/IndexedDB-2/#key-construct) that works on code units

This is I18N comment https://github.com/w3c/i18n-activity/issues/362. Please add the i18n-comment label.

inexorabletash:
DOMStrings are not UTF-16. They are sequences of 16-bit code units; they are not required to be valid UTF-16.

This is equivalent to calling `sort()` on an ECMAScript array of strings, which is the model for the behavior. I'll add a note.

inexorabletash:
PR at https://github.com/w3c/IndexedDB/pull/182

Also note that this is existing behavior from previous iterations of the spec.

aphillips:
I'm not quite buying it.  I mean, yes, they aren't _exactly_ UTF-16, in exactly the same way as we're discussing in the WHATWG infra space just now. But really they are strings and there is a clear relationship to Unicode. The sort you mention is fast and easy, but ignores supplementary characters and the unfortunate sorting of those characters a surrogate pairs. I18n would prefer to make supplementary characters work as first class citizens. And not that this is an index, not just a bag of strings. There is provision for range operations and these fail utterly on surrogate pairs.

inexorabletash:
"that this is an index, not just a bag of strings" 

The "sorted list" construct is used _only_ when getting the names of a set of object stores or indexes:

* [IDBDatabase.objectStoreNames](https://w3c.github.io/IndexedDB/#dom-idbdatabase-objectstorenames)
* [IDBObjectStore.indexNames](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-indexnames)
* [IDBTransaction.objectStoreNames](https://w3c.github.io/IndexedDB/#dom-idbtransaction-objectstorenames)

The names are identifiers specified by the developer in code.

And... this is unchanged since the first version of the spec (the latter is new, but must be consistent), and is interoperably deployed in all browsers already.

aphillips:
[...]

Issue #38, opened 2015-09-21
Add locale-aware sorting


annevk:
See https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#Locale-aware_sorting and in particular https://bugzilla.mozilla.org/show_bug.cgi?id=871846 for details.

@reuben can probably help out if there are any questions.

inexorabletash:
Yeah, I'd talked w/ bent about this. Thanks for filing it here.

The two big questions I had at the time were (1) stability of collations over time and (2) behavior of "auto" when device locale changes.

Copy/pasting from the old list: https://www.w3.org/2008/webapps/wiki/IndexedDatabaseFeatures

* Add 'locale' option to indexes (specified or 'platform')
* Add 'IDBLocaleAwareKeyRange' when querying such an index (doesn't enforce upper/lower order if both specified)
* Questions: are locales consistent across browsers? are collations stable?

annevk:
I suspect the answers are no and no, but @aphillips can probably give a clearer answer.

reuben:
No and no. Consistency depends on what versions of ICU browsers use and what locales they decide to ship. Collations change due to government regulations and new languages being introduced in the Unicode standard.

aphillips:
Locales are not necessarily going to be consistent across browsers. The locales available to a given browser depends on the platform it is running on and/or implementation choices by the browser vendor. Not all browsers use ICU internally, for example. And different locales may or may not be provisioned on different host environments.

CLDR and the Unicode Collation Algorithm (UCA) do provide a basis for building and tailoring collations. These do evolve over time as the data is updated, new characters are added to Unicode, and so forth. Classical database vendors (such as Oracle for example) often go to great lengths to preserve existing collations in a stable manner. This is a good thing for their customers, but may not be suitable for the Web platform. Note that there are ways to tailoring ICU's implementation, at least, to be consistent (by providing the collation rules).

Also, UCA and CLDR describe additional settings (which can appear in a locale identifier courtesy of the 'u' extension of BCP 47 (cf. RFC 6067 and UTR#35)) beyond those in what you might be thinking of as "the locale" and these are also provided in the JavaScript internationalization extension. The JS folks felt that collation stability or even consistency was less important than being as close to end-user expectations at possible.

sicking:
[...]

PR #182, opened 2017-03-23
Add note about sorting by code unit

inexorabletash:
For https://github.com/w3c/IndexedDB/issues/179

Add a note that sorting by code unit is for consistency with ECMAScript's Array.prototype.sort

### DIFF
--- a/index.html
+++ b/index.html
@@ -1850,8 +1850,6 @@ request<span class="p">.</span>onupgradeneeded <span class="o">=</span> <span cl
 from the database. Ideally the user will never see this.</p>
    </aside>
    <h2 class="heading settled" data-level="2" id="constructs"><span class="secno">2. </span><span class="content">Constructs</span><a class="self-link" href="#constructs"></a></h2>
-   <p>A <dfn class="dfn-paneled" data-dfn-type="dfn" data-noexport="" id="sorted-list">sorted list</dfn> is a list containing strings
-sorted in ascending order by code unit.</p>
    <p>A <dfn class="dfn-paneled" data-dfn-type="dfn" data-noexport="" id="name">name</dfn> is a string equivalent to a <code class="idl"><a data-link-type="idl" href="https://heycam.github.io/webidl/#idl-DOMString">DOMString</a></code>;
 that is, an arbitrary sequence of 16-bit code units of any length,
 including the empty string. <a data-link-type="dfn" href="#name" id="ref-for-name-1">Names</a> are always compared as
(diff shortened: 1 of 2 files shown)


inexorabletash:
@aphillips - can you take a look?

inexorabletash:
One thing I'd like to add to that is a note referencing something indicating that it's a poor choice for sorting user-visible string. Charmod-Norm doesn't seem to be the right reference - any suggestions?

inexorabletash:
... and/or note that a locale-aware sort could be done with:

`Array.from(db.objectStoreNames).sort(new Intl.Collator('fr').compare)`

index.html
```diff
@@ -1866,6 +1864,9 @@ <h2 class="heading settled" data-level="2" id="constructs"><span class="secno">2
   or something similar to map the provided name to a string that it
   can store.</p>
    </aside>
+   <p>A <dfn class="dfn-paneled" data-dfn-type="dfn" data-noexport="" id="sorted-list">sorted list</dfn> is a list containing strings
+sorted in ascending order by code unit.</p>
```
aphillips: Looking at the ED, this seems to use a "forward definition". To simplify this, I'd probably move this paragraph below the note about names. That way you gain the discussion of DOMString's 16-bit code unit container goodness and don't have to repeat yourself here. Otherwise I'd add "16-bit" to "code unit"?

inexorabletash: Totally agreed - I updated the ordering in the most recent update to this PR.

inexorabletash: ... and just added "16-bit", thanks!

aphillips:
> ... and/or note that a locale-aware sort could be done with:

@inexorabletash How does that work with cursors? ;-)

> One thing I'd like to add to that is a note referencing something indicating that it's a poor choice for sorting user-visible string. Charmod-Norm doesn't seem to be the right reference - any suggestions?

Maybe [UTS10](http://unicode.org/reports/tr10)? LTLI isn't super close to done and only talks obliquely about it. Maybe the INTL spec (ECMA-402)?

inexorabletash:
> How does that work with cursors? ;-)

For actual content (not the database schema) you're stuck with either pulling it into memory and sorting (as you noted elsewhere), or coming up with a more complex key scheme. For example, if you wanted to sort by code point you could encode the string as UTF-32 into an Uint32Array and use that as the key. Code point isn't actually useful, of course. Looking at UTS10 I see _sort key_ defined, which sounds intriguingly like the right thing, but we're going beyond what I know.

(Anyway, https://github.com/w3c/IndexedDB/issues/38 is the right place for this discussion.)

aphillips:
I'll take up the topic of language-aware and Unicode sorting in #38.

I don't think the "flesh out" edit is a good one as it goes in the wrong direction. It says how one could work around the fact that the sorted name list is in code point order rather than saying (as the section head suggests) why this order was chosen. I'd suggest:

> This matches the `Array.prototype.sort` on an `Array` of `Strings`. This ordering compares the 16-bit code units in each string, producing a highly efficient, consistent, and deterministic sort order. The resulting list will not match any particular alphabet or lexicographical order, particularly for code points represented by a surrogate pair.

inexorabletash:
Perfect, I like that much better!

PR #183, opened 2017-03-24
Define 'ascending' as alias for 'greater than'

inexorabletash:
Link 'ascending' to the definition for 'greater than' for keys, which provides the ordering.


inexorabletash:
@aphillips - can you take a look?

Issue #178, opened 2017-03-23
Add health warning for opaque string comparison


aphillips:
https://www.w3.org/TR/IndexedDB-2/#object-store-construct

> Object store names are always compared in a case-sensitive manner, as opaque sequences of 16-bit code units.

The object store name comparison is fine, so long as "comparison" means "for identity/equality". Case-sensitivity is _recommended_ by the I18N WG and [Charmod-Norm](http://w3c.github.io/charmod-norm/). However, a health warning may be warranted here, since it is more than just case-sensitive: it is sensitive to the exact encoding sequence. For example, normalization form affects this. Charmod-Norm provides a somewhat exhaustive list of this kind of variation ;-).

A health warning might be:

> Object store names are always compared as opaque sequences of 16-bit code units. As a result, name comparison is sensitive to variations in case as well as other minor variations such as normalization form, the inclusion or omission of controls, and other variations in Unicode text. [[Charmod-Norm]]

This is I18N comment https://github.com/w3c/i18n-activity/issues/363. Please add the i18n-comment label.

inexorabletash:
Note that databases, object stores, and indexes all have names which behave the same way. I'll spin up a construct for that.

inexorabletash:
PR: https://github.com/w3c/IndexedDB/pull/184

@aphillips - can you review?

(and in case I haven't said it yet - thanks for all the awesome feedback!)

Issue #177, opened 2017-03-23
Name content issues?


aphillips:
https://www.w3.org/TR/IndexedDB-2/#object-store-construct

> An object store has a name. The name can be any string value, including the empty string. 

Names are just strings of UTF-16 code units (that's what "any string value" means I think). There are no content restrictions on names. For example, a name could be 14 "shift in" control characters. Or it could consist of a non-character such as U+FFFE. Should "string value" be more closely defined?

This is I18N comment https://github.com/w3c/i18n-activity/issues/364. Please add the i18n-comment label

inexorabletash:
Names are anything which can be specified as a DOMString (WebIDL language) a.k.a. ECMAScript string. So "sequences of 16-bit code units". Valid UTF-16 is not required.

I'll tackle this along with https://github.com/w3c/IndexedDB/issues/178

inexorabletash:
PR: https://github.com/w3c/IndexedDB/pull/184

@aphillips - can you review?

aphillips:
I'm never thrilled by UTF-16 code unit sorting. Yes, sorting by code point is a pain in the ass. But it produces better results and it isn't *that* hard to do. However, I'm willing to hold my nose and won't reopen this issue. The described sort order is deterministic and fast and probably won't be seen by end users often.

PR #184, opened 2017-03-24
Define a 'name' construct, reference Charmod-Norm

inexorabletash:
Replace repeated "opaque sequences of 16-bit code units" blather with a type with associated details.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -369,6 +369,23 @@ from the database. Ideally the user will never see this.
 A <dfn>sorted list</dfn> is a list containing strings
 sorted in ascending order by code unit.
 
+A <dfn>name</dfn> is a string equivalent to a {{DOMString}};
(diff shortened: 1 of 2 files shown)


inexorabletash:
This is for #177 and #178 

@aphillips - can you review?

aphillips:
LGTM. Are nulls allowed in DOMString? I don't recall. If not, that should be called out.

inexorabletash:
"\u0000" is valid input when creating a name - so no special case. Unless you meant nullable string, in which case that'd be `DOMString?` in WebIDL parlance, but the inputs here are not nullable.

Tests here, BTW: 

https://github.com/w3c/web-platform-tests/blob/master/IndexedDB/string-list-ordering.htm

(Yes, even covers surrogate pairs being sorted by code unit.)

Issue #11, opened 2015-05-18
Validate exception ordering against implementations


inexorabletash:
The various "If .... throw ...." statements were left in order from previous iterations of the spec which asserted behavior but didn't define it as formal steps.

We need to verify that the steps match the implementations in the case of conflicts, and reorder if necessary. If implementations disagree, file implementation bugs as necessary.

> NOTE: Checkboxes below indicate that at least 2 implementations and the spec are in alignment.

inexorabletash:
IDBDatabase:
* createObjectStore:
 * FF: InvalidStateError > SyntaxError > ConstraintError > InvalidAccessError
 * Cr: InvalidStateError > TransactionInactiveError > ConstraintError > SyntaxError > InvalidAccessError
* deleteObjectStore
 * FF: InvalidStateError > NotFoundError
 * Cr: InvalidStateError > TransactionInactiveError > NotFoundError
* transaction
 * FF: InvalidStateError > InvalidAccessError (empty) > NotFoundError > InvalidAccessError (mode)
 * Cr: InvalidAccessError (empty) > InvalidAccessError (mode) > InvalidStateError > NotFoundError

inexorabletash:
IDBObjectStore:
* add/put:
 * FF: TransactionInactiveError > ReadOnlyError > DataError
 * Cr: InvalidStateError (deleted) > TransactionInactiveError > ReadOnlyError > DataError
* clear
 * FF: TransactionInactiveError > ReadOnlyError
 * Cr: InvalidStateError (deleted) > TransactionInactiveError > ReadOnlyError
* createIndex
 * FF: SyntaxError > InvalidStateError > ConstraintError > InvalidAccessError
 * Cr: InvalidStateError (deleted) > TransactionInactiveError > SyntaxError > ConstraintError > InvalidAccessError
* delete
 * FF: TransactionInactiveError > ReadOnly > DataError
 * Cr: InvalidStateError (deleted) > TransactionInactiveError > ReadOnly > DataError
* deleteIndex
 * FF: InvalidStateError > NotFoundError
 * Cr: InvalidStateError (deleted) > TransactionInactiveError > NotFoundError
* get/getAll/getAllKeys/count/openCursor/openKeyCursor
 * FF: TransactionInactiveError > DataError
 * Cr: InvalidStateError (deleted) > TransactionInactiveError > DataError

inexorabletash:
IDBIndex
------------
[...]

PR #188, opened 2017-04-04
Align IDBDatabase.transaction() exception precedence with tests/implementations

inexorabletash:
As noted in https://github.com/w3c/web-platform-tests/issues/5313 the spec disagreed with implementations and tests for exception precedence for IDBDatabase.transaction()

Per https://github.com/w3c/IndexedDB/issues/11:

InvalidStateError (closed/version) > InvalidAccessError (empty) > NotFoundError > TypeError (mode)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1179,7 +1179,7 @@ with [=transaction/cleanup event loop=] matching the current
 <aside class=note>
   This behavior is invoked by [[HTML]]. It ensures that
   [=/transactions=] created by a script call
-  to{{IDBDatabase/transaction()}} are deactivated once the task that
+  to {{IDBDatabase/transaction()}} are deactivated once the task that
   invoked the script has completed. The steps are run at most once for
   each [=/transaction=].
 </aside>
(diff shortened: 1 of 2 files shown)


inexorabletash:
@pwnall - can you review?

index.bs
```diff
@@ -2628,26 +2628,26 @@ instance on which it was called.
 The <dfn method for=IDBDatabase>transaction(|storeNames|,
 |mode|)</dfn> method, when invoked, must run these steps:
 
-1. If |mode| parameter is not {{"readonly"}} or {{"readwrite"}},
-    [=throw=] a [=TypeError=].
-
-2. If this method is called on {{IDBDatabase}} object for which an
+1. If this method is called on {{IDBDatabase}} object for which an
```
pwnall: on an {{IDBDatabase}}?

FWIW, I liked instance a bit better than object. Also, I think you can get away without either.

index.bs
```diff
@@ -2628,26 +2628,26 @@ instance on which it was called.
...
 
-5. If any string in |scope| is not the name of an [=/object
+4. If any string in |scope| is not the name of an [=/object
     store=] in the [=connected=] [=database=], [=throw=] a
     "{{NotFoundError}}" {{DOMException}}.
 
-6. If |scope| is empty, [=throw=] an "{{InvalidAccessError}}" {{DOMException}}.
+5. If |scope| is empty, [=throw=] an "{{InvalidAccessError}}" {{DOMException}}.
+
+6. If |mode| parameter is not {{"readonly"}} or {{"readwrite"}},
```
pwnall: the |mode| parameter, or just |mode|?

inexorabletash: Just |mode| - thanks!

index.bs
inexorabletash: I went with "the connection" instead.

PR #199, opened 2017-05-05
Editorial: Remove use of 'substeps', simplify wording, add 'in parall…

inexorabletash:
…el' links

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2235,56 +2235,57 @@ when invoked, must run these steps:
 
 4. Let |request| be a new [=open request=].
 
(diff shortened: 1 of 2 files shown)


brettz9:
In the last commit, starting with the substep "Wait until all previously added [=requests=] in |transaction|", it looks like the numbering should start over at 1.

inexorabletash:
Good catch!

inexorabletash:
@brettz9 - can you give the latest another look?

brettz9:
All LGTM!

PR #201, opened 2017-05-11
Define db's associated upgrade tx to align with impls/tests

### LINKED ISSUE #192: create/deleteObjectStore: InvalidStateError logic does not match impls/tests
Spec has:

> Let transaction be the upgrade transaction associated with database. If one does not exist or it is finished, throw an "InvalidStateError" DOMException.

But in this case, a TransactionInactiveError is thrown instead in Chrome/Firefox and asserted as such in tests:

```js
tx.abort();
db.createObjectStore('s'); // throws TransactionInactiveError
```

Noted by the ever vigilant @brettz9 over in https://github.com/w3c/web-platform-tests/issues/5612 but it's the spec that's wrong here since impls/tests match.

Unfortunately, the "finished" notion for transactions is a bit squishy. Spec-wise we may be able to just drop the "or it is finished" bit from these methods and then tidy up the transaction lifecycle at some point in the future.

inexorabletash:
Implementations and tests have create/deleteObjectStore throw "TransactionInactiveError" if the transaction has aborted but the abort event has not yet fired. The spec disagreed, or at least was imprecise about the definition of "finished". 

To resolve this, databases are given an associated upgrade tx which is nulled out when the event fires, rather than relying on "finished". This simplifies text elsewhere too.

@brettz9 - can you review this change? This should address the outstanding issue you noted over in https://github.com/w3c/web-platform-tests/issues/5612

Tracking bug for this is #192

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7163,6 +7163,11 @@ document's Revision History</a>.
     the [=asynchronously execute a request=] steps.
     (<a href="https://github.com/w3c/IndexedDB/issues/192">issue #192</a>)
 
+* Use [[HTML]]'s <a abstract-op>StructuredSerializeForStorage</a>
+    hook.
+    (<a href="https://github.com/w3c/IndexedDB/issues/197">issue #197</a>,
+     <a href="https://github.com/w3c/IndexedDB/issues/197">issue #152</a>)
+
 * Define [=database=]'s associated [=database/upgrade transaction=] to
     align exceptions thrown from {{IDBDatabase/createObjectStore()}} and
     {{IDBDatabase/deleteObjectStore()}} with tests and implementations.
--- a/index.bs
+++ b/index.bs
@@ -7163,10 +7163,10 @@ document's Revision History</a>.
     the [=asynchronously execute a request=] steps.
     (<a href="https://github.com/w3c/IndexedDB/issues/192">issue #192</a>)
 
-* Use [[HTML]]'s <a abstract-op>StructuredSerializeForStorage</a>
-    hook.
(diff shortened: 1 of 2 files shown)


inexorabletash:
Testing for this is already covered in WPT:

IndexedDB/idbdatabase-createObjectStore-exception-order.htm
IndexedDB/idbdatabase-deleteObjectStore-exception-order.htm

brettz9:
While I'm still checking (looks good so far), I did have the following concern/question... It is admittedly a concern about a missing test rather than spec issue, but I also wanted to confirm what the testing behavior would be if it existed...

`idbtransaction-objectStore-finished.html`, and `idbobjectstore-index-finished.html` cover the expected `InvalidStateError` throwing immediately after an `abort` (and suggest by their test names that "finished" at least definitively occurs within `abort`). However, I don't see any tests for an `abort` throwing immediately after another `abort` (as would be consistent with this understanding (of one subset) of "finished" since `abort` is to throw upon a transaction being "finished"). (FWIW, besides my manual checking, my regex to check the files was `abort\([\s\S]*abort\(`.) Should there be such a test which checks for `InvalidStateError` immediately after a prior `abort`?

brettz9:
Besides my somewhat-related question about the expected behavior of multiple aborts, the PR looks good.

inexorabletash:
> I don't see any tests for an abort throwing immediately after another abort (as would be consistent with this understanding (of one subset) of "finished" since abort is to throw upon a transaction being "finished").

Yes, it does look we're missing a test for that case.

Issue #161, opened 2017-03-03
The done flag is never unset between upgradeneeded and success


inexorabletash:
There's a note in the [request definition](https://w3c.github.io/IndexedDB/#request-construct):

> Requests are not typically re-used, but there are exceptions. When a cursor is iterated, the success of the iteration is reported on the same request object used to open the cursor. And when an upgrade transaction is necessary, the same open request is used for both the upgradeneeded event and final result of the open operation itself. In both cases, the request’s done flag will be unset then set again, and the result may change.

For cursor iteration there are explicit _"Unset the done flag on request."_ steps (e.g. in [continue()](https://w3c.github.io/IndexedDB/#dom-idbcursor-continue)). But there's nothing in the [steps for upgrade transactions](https://w3c.github.io/IndexedDB/#upgrade-transaction-steps).

I'd expect a final step there that unsets the done flag. Ostensibly this would be visible as a change in the request's `readyState` between the "complete" event fired at the transaction and the "success" event fired at the request. 

However... I'm unable to test this in either Chrome or Firefox. In Chrome, at least, the code that queues the "complete" and "success" doesn't really leave room for anything to happen in between, at least not in a non-racy way.

So... the note is inconsistent with the spec, but changing the spec without a test seems weird.

brettz9:
Also, if there is an error (abort errors are the specified ones) during or after the "steps to run an upgrade transaction", should `open`  not reset the `result` to `undefined` as well? Or can an error co-exist with a result?

Also, as far as the explanatory text you cited mentioning the `result` possibly changing... If in `IDBFactory.open`, the "steps to open a database"  are run *in parallel* with the queuing of a task to set the `result` or `error`, I would think this task to set a result for open would technically occur before the task to set a result for the upgrade (within the "steps to run an upgrade transaction"). While the open steps could clarify that the steps should wait until there is a result, the task would still have already been added to the queue before the upgrade task had been added to it.

brettz9:
Also, since the open request is not using the steps to asynchronously execute a request, the `done` flag is also not explicitly *set* anywhere upon a success (though the general definition for requests in general does state "If a request completes successfully, the done flag is set").

inexorabletash:
Thanks as always @brettz9 !

I tacked this in https://github.com/w3c/IndexedDB/pull/202

> I'd expect a final step there that unsets the done flag.

I added this in the case where the result is cleared out for abort. I still don't think it's observable, but it made sense to do as part of the related changes...

[...]

PR #202, opened 2017-05-11
Set request result/done flag appropriately in open/delete

inexorabletash:
For https://github.com/w3c/IndexedDB/issues/161:

* When an open or delete succeeds the done flag should be set.
* When an open fails aborts, the request's result should be cleared.

I need to verify we have WPT coverage for this.

@brettz9 - can you review?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1273,9 +1273,9 @@ A [=request=]'s [=get the parent=] algorithm returns the request's
   on the same [=request=] object used to open the cursor. And when
   an [=/upgrade transaction=] is necessary, the same [=open
   request=] is used for both the <code>[=upgradeneeded=]</code>
-  event and final result of the open operation itself. In both cases,
+  event and final result of the open operation itself. In some cases,
   the request's [=request/done flag=] will be unset then set again, and the
-  [=request/result=] can change.
+  [=request/result=] can change or [=request/error=] could be set insead.
 </aside>
 
 <!-- ============================================================ -->
@@ -2258,6 +2258,7 @@ when invoked, must run these steps:
     2. [=Queue a task=] to run these steps:
 
         1. If |result| is an error,
+            set |request|'s [=request/result=] to undefined,
             set |request|'s [=request/error=] to |result|,
             set |request|'s [=request/done flag=],
             and [=fire an event=] named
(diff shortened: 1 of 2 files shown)


inexorabletash:
WPT coverage was lacking! Fortunately, Chrome and Firefox already pass.

New tests: https://github.com/w3c/web-platform-tests/pull/5899

brettz9:
I think this PR covers it well...

However, besides if you wanted to add that invisible step to unset the done flag for successful transactions, there is the case also of `result` always needing to return `undefined` (per the `result` getter) when there is an error, so I don't know if you wanted to make that explicit for `open` and `deleteDatabase`.

For example, for the following in the steps to open a database, "If connection was closed, create and return a newly created "AbortError" DOMException and abort these steps.", this isn't followed by setting the `result` to `undefined` (as now is the case with your PR if the steps to abort are triggered). And perhaps this could also occur with other errors of the database (like `UnknownError`) that didn't cause an abort.

I think mention of setting the request result to `undefined` could be added close to the end of `open()` and `deleteDatabase()` when the request's `error` properties are set.

inexorabletash:
Thanks!

> ...if you wanted to add that invisible step to unset the done flag..

Rather than doing this I updated the non-normative comment so it doesn't claim the done flag will be unset in all cases.

> ...If connection was closed... this isn't followed by setting the result to undefined

Great catch! I addressed this as you suggested...

> .. setting the request result to undefined could be added close to the end of open() and deleteDatabase() when the request's error properties are set.

... but only for open(), since deleteDatabase() doesn't have a path where the request would be "done" and get a result/error except at the very end.

brettz9:
> ... but only for open(), since deleteDatabase() doesn't have a path where the request would be "done" and get a result/error except at the very end.

Oh yes... I c...there will be nothing to clear...

So considering the `result` getter's definition:

> The result attribute’s getter must throw an "InvalidStateError" DOMException if the done flag is unset. Otherwise, the attribute’s getter must return the result of the request, or undefined if the request resulted in an error.

...upon a `deleteDatabase` error, there will be no `InvalidStateError ` error on retrieval of the `result` property as the done flag has been set by the end of `deleteDatabase`, and the getter is already spec'd to return `undefined` if there is an error, so there is no need to explicitly set it (with nothing prior on the delete request to overwrite)...

inexorabletash:
Yes. It could be made a bit more rigorous, as "if the request resulted in an error" is bit of action-at-a-distance. But for this CL... everything look good?

brettz9:
Yup, all good, thanks!

Issue #207, opened 2017-06-08
Implementor's note: Escape the database name?


jcjones:
I was reviewing the current editor's draft for security issues and didn't find anything substantial (yay). That said, you may want to add an implementation note:

The algorithm for opening a database leaves unspecified how a user agent should store the database, but permits an arbitrary database name. Implementations should be careful that the database name, if converted into some kind of disk path, be appropriately escaped so an adversary can't use `../` or similar to violate the same-origin policy.

inexorabletash:
PR at https://github.com/w3c/IndexedDB/pull/210 - just added minimal text. Could add more about escaping other names (e.g. if writing to databases), or if e.g. hashing origins to avoid collisions, or being cautious about truncation. Additional verbiage welcome!

brettz9:
These might not translate to any security problems, but in our Node implementation (which relies on sqlite3 that does convert names to disk paths), we applied the following config for escaping file names for writing to disk (mostly to avoid name clashes):

1. Ensure an empty string database name is not supplied without modification as a file name
2. Requiring escaping of NFD expanding characters to avoid NFD normalization clashes (given NFD being performed on file names in MacOS). (Maybe not a feature for some!)
3. Escaping characters commonly still reserved on modern file systems (our JavaScript regex is `[\u0000-\u001F\u007F"*/:<>?\\|]`)
4. Requiring config not to throw if 255+ characters are used in a name given file name length limits.
5. Escape upper case ASCII characters which are not distinguished on some file systems (though use of alternate cases for other Unicode characters does not seem to be a problem from my limited testing, so we do not escape case for other Unicode characters).
6. Escape unmatched surrogates

inexorabletash:
Ping @jcjones - can you take a look at #210 ?

jcjones:
Thanks for the re-ping! I commented on #210 that it looks good to me, thanks!

inexorabletash:
Thanks! Closing this out.

@brettz9 - good enumeration of some of the implementation risks and mitigations. For now I'm leaving those out, as they're ostensibly covered by the existing non-normative note "If an implementation uses a storage mechanism which does not support arbitrary strings, the implementation can use an escaping mechanism or something similar to map the provided name to a string that it can store." although maybe it could use the word "safely" in there somewhere.

rjkmurray:
[...]

PR #210, opened 2017-06-22
Add note about escaping names

inexorabletash:
@jcjones - can you take a look? This is for https://github.com/w3c/IndexedDB/issues/207


jcjones:
Looks good to me! Sorry for the delay, thanks for the re-ping.

rjkmurray:
Thanks I followed all parameters it will save on redundancy issues that is a concern for all of us

PR #218, opened 2017-10-09
Fix several issue URLs

xfq:


inexorabletash:
Thanks for catching these!

PR #219, opened 2017-10-19
Validate HTML

xfq:
https://validator.w3.org/nu/?doc=https%3A%2F%2Fw3c.github.io%2FIndexedDB%2F :

> Element `details` is missing a required instance of child element `summary`.

I can adjust the text in the `summary` element if needed.

/cc @inexorabletash


inexorabletash:
Thanks! ("Details" is fine, can't think of anything significantly better)

Issue #226, opened 2018-01-15
Web IDL syntax error?


xfq:
When generating the spec with the latest Bikeshed[1], a fatal error is emitted:

```
FATAL ERROR: IDL SYNTAX ERROR LINE: 17 - skipped: "boolean includes(any key)"
 ✘  Did not generate, due to fatal errors
```

Not sure if it's a bug in the IDL fragment, Bikeshed, or widlparser.

-----

[1] Bikeshed commit: [d1c8ac9ac0bb0f0e970e3942633cfb7d2f7e12e6](https://github.com/tabatkins/bikeshed/commit/d1c8ac9ac0bb0f0e970e3942633cfb7d2f7e12e6)

xfq:
@inexorabletash @tabatkins

inexorabletash:
I don't see an obvious issue with the IDL fragment.

inexorabletash:
It's specific to the word "includes" - `includes` is a keyword in WebIDL, so it's probably making the parser sad.

tabatkins:
Yes, exactly. You can write `boolean _includes(any key)` instead - the leading `_` is stripped from the identifier, and functions as an escaping mechanism.

(Once you switch to using a `_` prefix, Bikeshed should handle things for you automatically; `{{includes()}}` should work fine.)

inexorabletash:
Cool. Just found that part of the WebIDL spec update too. :)

inexorabletash:
Pushed 98b28b9890eb92b36eaadbedb31cea2d3bcd64aa for this... @xfq will that auto-build and deploy ?

inexorabletash:
Looks like it worked. Thanks @xfq and @tabatkins !

PR #228, opened 2018-01-17
Fix the link to the first edition

xfq:


xfq:
Squashed and merged. The change won't be reflected in the generated Editor's Draft for the moment, due to #226.

Issue #230, opened 2018-02-15
Track ES BigInt work


inexorabletash:
As raised over in https://github.com/tc39/proposal-bigint/issues/123 

When BigInt support lands in ECMAScript they won't be usable out of the box as keys in Indexed DB.

inexorabletash:
If we add new key types we should toss boolean in as well. https://github.com/w3c/IndexedDB/issues/76

littledan:
Serialization of BigInts is proposed [in this HTML PR](https://github.com/whatwg/html/pull/3480). This would mean you can use BigInts and BigInt wrappers as IndexedDB values. Any concerns with this proposal?

For keys, I like @inexorabletash 's suggestions in https://github.com/tc39/proposal-bigint/issues/123 . To make it into a concrete straw-person--how does this plan sound?
- Permit BigInt primitive keys in IndexedDB
- Don't associate BigInt and Number keys with each other, even if they are mathematically equal, and always preserve the original type
- Make BigInts all come after any Number in sort order

Is there any particular need to make `autoIncrement` tie in with BigInts? Seems like the current semantics are pretty closely tied to Number, so this could be rather complex to work out. Is anyone running into issues with overflow of these counters?

aliams:
I just gave this a read and I think it all makes sense. It's possible that there could be an impact for some underlying databases by increasing the index size, but I suppose there are always way to deal with that.

littledan:
Wow, thanks for the extremely fast turnaround for a technical review from the editors! Impressive.

> Make BigInts all come after any Number in sort order, and before other types which follow Number.

Not sure how unambiguous this sentence is; I was imagining that BigInt would compare as less than Number. (Seems like there's nothing less than Number, so not sure what I meant by the second part.) However, I saw a note in the specification that made this sound like a bad idea:

> As a result of the above rules, negative infinity is the lowest possible value for a key. 

Seems like it might be good to preserve this property, in case someone is depending on it, and instead have all BigInts compare as greater than all Numbers.

aliams:
Sounds good to me!

inexorabletash:
**TPAC 2019 Web Apps Indexed DB triage notes:**

@littledan - have you heard any requests for this?

fosskers:
> If I manually cast down to a 64-bit float before writing, it works (but I've lost some bits of accuracy).

[...]

PR #231, opened 2018-02-19
Allow BigInts as IndexedDB keys

littledan:
Design decisions:
- BigInts compare greater than Number and less than everything else
- No type coercion between BigInt and Number; they are simply
  distinct, unrelated keys, even if mathematically equal
- No BigInt autoIncrement support--53 bits should be enough for anyone

BigInts and BigInt wrappers are proposed to be made serializable, and
therefore available as IndexedDB values, in
https://github.com/whatwg/html/pull/3480

Addresses #230

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -25,6 +25,17 @@ Complain About: accidental-2119 yes
 spec:html; type:dfn; for:/; text:task queue
 </pre>
 
+<pre class='biblio'>
+{
+  "BIGINT": {
+    "href": "https://tc39.github.io/proposal-bigint/",
+    "title": "BigInt Specification",
+    "publisher": "TC39",
+    "status": "Stage 3 proposal"
+  }
+}
+</pre>
+
 <pre class=anchors>
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: dom.html
@@ -63,6 +74,9 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
     type: dfn
         text: sequence<DOMString>; url: idl-sequence
         text: sequence<any>; url: idl-sequence
+spec: bigint; urlPrefix: https://tc39.github.io/proposal-bigint/
+    type: dfn
+        text: BigInt; url: #sec-ecmascript-language-types-bigint-type
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -622,6 +637,7 @@ be either:
 an {{unrestricted double}} if type is <i>number</i> or <i>date</i>,
 a {{DOMString}} if type is <i>string</i>,
 a list of {{octet}}s if type is <i>binary</i>,
+an integer if the type is <i>bigint</i>,
```
littledan: This line is a little fudge--the other types have meanings that are more well-defined in terms of Infra or WebIDL, but there isn't a BigInt WebIDL type yet.

inexorabletash:
Thanks for the PR!

To set expectations:

• We haven't kicked off 3.0 yet, and this should probably block on that. 

• We'd like one implementation (even nightly build/behind a flag) before merging.

• Need web-platform-test coverage, ideally landed first.

littledan:
OK, I don't have time to implement this right now, and I'm not sure if any implementers are interested. If anyone is interested, please respond here and I can write some web-platform-tests. Otherwise, we can consider this a fun thought exercise.

littledan:
It's rather circumstantial evidence, but I'm seeing not so much expression of people saying they would actually use BigInt as IndexedDB keys in [this Twitter poll](https://twitter.com/littledan/status/968094360960159744). By contrast, there was [a lot more interest](https://twitter.com/littledan/status/942817378009743363) in BigInt64 TypedArrays.

kaizhu256:
i'm just conjecturing, if the twitter poll was re-worded "Should *BigInt64* be supported as an IndexedDB key?" it may get the same enthusiasm as BigInt64 TypedArrays.

also keep in mind the potential impact to indexeddb from introduction of future javascript data-types under discussion (e.g.  bigdecimal and complex-numbers).

annevk:
@inexorabletash @asutherland any reason we're not doing this? This isn't blocked on IDL right as you're using `any` for these operations?

asutherland:
It's a reasonable enhancement for completeness, but adding new type support to keys implies potentially non-trivial enhancements to whatever the implementation is using for comparator/ordering purposes.  Given https://github.com/w3c/IndexedDB/pull/231#issuecomment-368863760 this does seem like a high cost/benefit ratio compared to other implementation and spec concerns.

inexorabletash:
FWIW, we've had no requests come through our bug channels or from partners.

annevk:
I'm going to close this then for clarity. Hope that's okay.

PR #232, opened 2018-03-15
Make "cleanup IDB transactions" return a value

domenic:
It tells you whether any cleanup work was performed or not. This helps with https://github.com/w3c/requestidlecallback/issues/70 and https://github.com/whatwg/html/pull/3570.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1173,15 +1173,21 @@ The following constraints define when a [=/transaction=] can be
 
 </div>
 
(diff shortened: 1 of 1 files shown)


inexorabletash:
Thanks! I just sent mail to kick off the "3.0" version of the spec; assuming that doesn't objections I'll rev the spec metadata and we can merge this.

PR #236, opened 2018-04-11
Use (DOMString or sequence<DOMString>) for keyPath

### LINKED ISSUE #235: Shouldn't keyPath properties be (DOMString or sequence<DOMString>)?
`keyPath` for IDBObjectStore and IDBIndex is defined as `any` but then the spec says:

>The key path is converted as a DOMString (if a string) or a sequence<DOMString> (if a list of strings), per [WEBIDL].

I think it should be `DOMString or sequence<DOMString>` then.

saschanaz:
Fixes #235 

[Per the spec](https://www.w3.org/TR/IndexedDB-2/#dom-idbobjectstore-keypath):

>The key path is converted as a DOMString (if a string) or a sequence<DOMString> (if a list of strings), per [WEBIDL].

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2721,7 +2721,7 @@ interface represents an [=/object store handle=].
 [Exposed=(Window,Worker)]
 interface IDBObjectStore {
   attribute DOMString name;
-  readonly attribute any keyPath;
+  readonly attribute (DOMString or sequence&lt;DOMString&gt;) keyPath;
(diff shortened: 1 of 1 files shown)


saschanaz:
Marked as non substantive for IPR from ash-nazg.

inexorabletash:
See https://github.com/w3c/IndexedDB/issues/235 - this would be invalid Web IDL.

PR #238, opened 2018-04-27
WindowOrWorkerGlobalScope is a mixin

saschanaz:
...and so extensions should use [`partial interface mixin`](https://heycam.github.io/webidl/#partial-interface-mixin) rather than `interface mixin`.

See also https://github.com/whatwg/html/pull/3650

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2165,7 +2165,7 @@ interface is present in the global scope of environments that support
 Indexed DB operations.
 
 <pre class=idl>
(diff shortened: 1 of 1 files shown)


inexorabletash:
Marked as non substantive for IPR from ash-nazg.

inexorabletash:
Thank you!

We'll want to update the tests for this as well - how do you feel about a PR for https://github.com/w3c/web-platform-tests/blob/master/interfaces/IndexedDB.idl ?

Issue #31, opened 2015-08-10
API to enumerate databases


inexorabletash:
Moved from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=16137

Discussions:

http://lists.w3.org/Archives/Public/public-webapps/2011JulSep/1528.html
http://lists.w3.org/Archives/Public/public-webapps/2011JulSep/1537.html

Critical bit: how do we do this such that the answer isn't obsolete as soon as it's given, since there are no transactions over the set-of-databases.

inexorabletash:
FWIW, based on early feedback when IDB was shiny and new, Chrome implemented/shipped a prefixed API for this:
```webidl
partial interface IDBFactory {
  IDBRequest webkitGetDatabaseNames();
};
```
Result is a `DOMStringList` (see #28) of the names, purely a snapshot at the time the request is processed.

aliams:
I think we should keep this around given that it's useful for database libraries and diagnostic purposes. I recommend that we pursue a standardized non-prefixed version of this API e.g. `getDatabaseNames()`.

Gilead:
I reckon no attempt needs to be made to ensure the results remain valid after the function call returns values. I would suggest leaving it up to the application author to deal with it.

That said, a separate function like hasDatabase(name) would be useful, with the proviso that other open/create requests for the same name would be queued until after the current code finishes dealing with IndexedDB (I'm not familiar with the implementation enough to suggest a specific mechanism for this).

pwnall:
@Gilead The only mechanism I can think of would be to hold onto a lock for the database, which is what you get with `indexedDB.open`.

aliams:
It could be treated just like another transaction (such as versionchange, readwrite, readonly) in the sense that no other transactions can execute while `getDatabaseNames` is running.

pwnall:
@aliams Sorry if I misunderstood your suggestion, but merely serializing the execution of `getDatabaseNames` doesn't seem sufficient to me. Another transaction could sneak in between the `getDatabaseNames` transaction and a subsequent `indexedDB.open` transaction. This results in an unintuitive behavior where `indexedDB.open` would error out when given a database returned by `getDatabaseNames`, because a delete transaction snuck between `getDatabaseNames` and `indexedDB.open`.

aliams:
[...]

PR #240, opened 2018-05-18
Add IDBFactory/databases()

inexorabletash:
As discussed in https://github.com/w3c/IndexedDB/issues/31

Needs tests and an implementation.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/240.html" title="Last updated on Sep 17, 2018, 10:14 PM GMT (5e67186)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/240/662963b...5e67186.html" title="Last updated on Sep 17, 2018, 10:14 PM GMT (5e67186)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -17,13 +17,13 @@ Editor: Ali Alabbas, Microsoft Corp. https://microsoft.com, alia@microsoft.com
 Editor: Joshua Bell, Google Inc. https://google.com, jsbell@google.com
 Group: webplatform
 Test Suite: https://github.com/web-platform-tests/wpt/tree/master/IndexedDB
+Favicon: logo-db.png
 Complain About: accidental-2119 yes
 Markup Shorthands: css no, markdown yes
 </pre>
 
 <pre class=link-defaults>
 spec:html; type:dfn; for:/; text:task queue
-spec:infra; type:dfn; text:list
 </pre>
 
 <pre class=anchors>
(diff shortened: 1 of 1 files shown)


inexorabletash:
Thanks! Again, will wait for tests and an impl before pushing forward. Just wanted to kick the tires a bit.

>  I am still on the fence about the value of having the version information. 

Me too, but it doesn't seem to hurt too much, and if we just returned a list of names it would be harder to expand the results in the future.

If we didn't include versions it would encourage developers to open a database to reason further about it, and then they avoid doing more racy work, but being too opinionated in the past hasn't worked out well.

rjfioravanti:
Gmail already has use cases where we open the database to check the current version. It would be good for us if the version info was included.

foolip:
@pwnall, would you mind filing implementation bugs for this, pointing to the tests?

index.bs
```diff
@@ -2344,6 +2363,42 @@ when invoked, must run these steps:
...
+
+The <dfn method for=IDBFactory>databases()</dfn> method,
+when invoked, must run these steps:
+
+1. Let |origin| be the [=/origin=] of the global scope used
+    to access this {{IDBFactory}}.
+
+1. If |origin| is an [=opaque origin=],
+    then return a new promise rejected with a "{{NotSupportedError}}" {{DOMException}}.
+
```
brettz9: The other `IDBFactory` methods (`open` and `deleteDatabase`) throw a `SecurityError` DOMException upon an opaque origin. Is this deviation here deliberate?

brettz9: Also, I see no indication, unlike the other `IDBFactory` methods, of what should happen if there are any errors in retrieving the info... If this would occur as a rejected Promise, does the error format need to be documented?

Issue #234, opened 2018-04-10
Proposal: Add "writeonly" transaction mode


inexorabletash:
Forked off from #34 and previous mentioned on the [old wiki]( https://www.w3.org/2008/webapps/wiki/IndexedDatabaseFeatures), but formalizing it a bit:

* Introduce a new [transaction mode](https://w3c.github.io/IndexedDB/#transaction-mode), "writeonly"
* If a transaction's mode is "writeonly":
  * any IDBObjectStore or IDBIndex operation other than `put()` or `add()` will throw `NotSupportedError` 
  * `put()` and `add()` return null (return type changed to `IDBRequest?`)
  * `success`/`error` events not fired (i.e. don't bubble up to the transaction/connection)
  * a failed operation (usually, conflicting `add()`) still causes the transaction to abort
  * `complete`/`abort` events are still fired against the transaction

There is probably additional spec text required for clarity, but effectively since the transaction can no longer become active after the [cleanup transactions](https://w3c.github.io/IndexedDB/#cleanup-indexed-database-transactions) steps run when the task ends, the transaction can attempt to commit immediately; it needs to wait until all of the `put()`s/`add()`s finish. The [transaction lifetime](https://w3c.github.io/IndexedDB/#transaction-lifetime-concept) steps are still a bit wishy-washy; I didn't completely revise those as part of the big 2.0 rework of the spec.

inexorabletash:
The big use case for this is performance:

* general throughput - waiting for all of the success events to round-trip before committing (even if they are no-ops) can be a bottleneck
* fire-and-forget - during shutdown/navigation/suspension it is desirable to attempt to persist data without requiring additional round-trips and stalling such a process

inexorabletash:
@aliams, @wanderview, @beidson - any initial feedback (not official opinions) on this idea? A non-starter, or something that would be worth the Chrome team prototyping?

wanderview:
301 @asutherland

inexorabletash:
FYI, I sketched out another idea a few years ago - adding an explicit `commit()` method to the transaction that would immediately and permanently deactivate it to prevent further requests from being filed. [gist](https://gist.github.com/inexorabletash/d55a6669a040e92e47c6) - this might be more or less appealing than a distinct mode. The down side is that success/error events are still fired against each request even if the transaction is not activated on their callbacks, and that's still unnecessary overhead in some cases.

asutherland:
[...]

PR #242, opened 2018-06-18
Add IDBTransaction commit() method

inexorabletash:
Per discussion in https://github.com/w3c/IndexedDB/issues/234 - add an explicit `commit()` method to IDBTransaction.

* Reify an internal "state" for transaction (active, inactive, committing, finished) replacing the active flag.
* Make transaction life-cycle somewhat more rigorous.
* Add "processed" state to request, distinct from "done" (when the result has been delivered).
* Firm up the notion of a transaction's request list w/r/t processing state of requests.
* Firm up commit/abort steps w/r/t processing state of requests.
* Add commit() method itself.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -976,15 +976,10 @@ following:
   </dd>
 </dl>
 
-A [=/transaction=] has an <dfn>active flag</dfn>, which determines
-if new [=requests=] can be made against the transaction. A
-transaction is said to be <dfn>active</dfn> if its [=transaction/active flag=]
-is set.
-
 A [=/transaction=] optionally has a <dfn>cleanup event loop</dfn>
 which is an [=event loop=].
 
-A [=/transaction=] has a <dfn>request list</dfn> of [=requests=]
+A [=/transaction=] has a <dfn>request list</dfn> of pending [=requests=]
 which have been made against the transaction.
 
 A [=/transaction=] has a <dfn>error</dfn> which is set if the
@@ -1001,8 +996,57 @@ is a [=/transaction=] with [=transaction/mode=] {{"readwrite"}}.
 
 
 <!-- ============================================================ -->
-### Transaction Lifetime ### {#transaction-lifetime-concept}
+### Transaction Lifecycle ### {#transaction-lifecycle}
 <!-- ============================================================ -->
+
+A [=/transaction=] has a <dfn>state</dfn>, which is one
+of the following:
+
+<dl>
+  <dt><dfn>active</dfn>
+  <dd>
+    A transaction is in this state when it is first [=transaction/created=],
+    and during dispatch of an event from a [=/request=] associated with the transaction.
+
+    New [=/requests=] can be made against the transaction when it is in this state.
+
+  <dt><dfn>inactive</dfn>
+  <dd>
+    A transaction is in this state after control returns to the event
+    loop after its creation, and when events are not being dispatched.
+
+    No [=/requests=] can be made against the transaction when it is in this state.
+
+  <dt><dfn>committing</dfn>
+  <dd>
+    Once all [=/requests=] associated with a transaction have completed,
+    the transaction will enter this state as it attempts to [=commit=].
+
+    No [=/requests=] can be made against the transaction when it is in this state.
+
+  <dt><dfn>finished</dfn>
+  <dd>
+    Once a transaction has committed or aborted, it enters this state.
+
+    No [=/requests=] can be made against the transaction when it is in this state.
+</dl>
+
+<aside class=note>
+
+The lifecycle of a transaction can be visualized according the following
+diagram:
+
+<pre class='railroad'>
+OneOrMore:
+  Seq:
+    T: active
+    T: inactive
+Choice:
+  T: committing
+  N: abort
+T: finished
+</pre>
+</aside>
 
 Transactions are expected to be short lived. This is encouraged by the
 <a for=transaction lt=commit>automatic committing</a> functionality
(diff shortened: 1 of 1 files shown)


inexorabletash:
cc: @spanicker and @fmeawad - your input on this CL would be welcome. Note that it's mostly refactor to support the addition.

brettz9:
Just a nit observation--some of the new text creates lines going past the usual 80 character limit...

asutherland:
I think there's a hole or two in the spec around the "committing" state and aborting, at least as I understood the intent of the commit() changes to be.

Previously, the spec abstracted away most of the realities of the IDB logic running on some other event loop that's asynchronously accessed.  A transaction being "finished" was accordingly posed as whether "commit a transaction" or "abort a transaction" had run.  "Commit a transaction" now surfaces this reality by introducing the "committing" state so that new requests can't be issued and to more accurately express that the already-enqueued requests could still fail and cause the transaction to abort.

The issues I see with the patch as stands are:
- abort() post-patch only checks for a state of transaction/finished before throwing.  This creates an awkward semantic state where `commit(); abort();` could be invoked and neither method would throw because abort() is okay with being invoked with the transaction in the "committing" state, but either one of the two results could happen depending on the implementation.
- Similarly, it seems like "fire a success event" can still trigger an (ambiguous) abort even after commit() is called.  It directly invokes the "abort a transaction" algorithm without any guards, but I think a transaction state check of [committing, finished] is appropriate.  ("Fire an error event" does this too, but we would expect the abort to have been internally triggered in that case already?).
- commit() should probably mention the impact it will have on "success" events if we are going with what I understood.  So we'd explicitly say that they will still fire but if they throw, the transaction will not be aborted.  Also we'd possibly say that the transaction will still commit even if the events are unable to be fired.  For example `transaction.commit(); self.close();` in a worker should still cause the transaction to commit.
- transaction/lifetime should also mention the "committing".  Alternately, the "finished" state could perhaps instead be "completed" and the original semantics of "finished" meaning "abort a transaction" or "commit a transaction" having been invoked (at least once) could be restored.

index.bs
```diff
@@ -3142,7 +3150,7 @@ must run these steps:
 1. If |store| has been deleted, [=throw=] an
     "{{InvalidStateError}}" {{DOMException}}.
 
-1. If |transaction| is not [=transaction/active=], [=throw=] a
+4. If |transaction| is not [=transaction/active=], [=throw=] a
```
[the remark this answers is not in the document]
inexorabletash: Thanks, fixed in 0b9c57bcc1d0df0ecaf7624fe52b60ef1b4ff97a

Re: state reference implicit - any thoughts on changing the non-algorithm transaction/active references? (There appear to be 11 or so of them) It should probably have a separate term and definition. Maybe "a transaction is said to be _accepting requests_ if its _state_ is _active_" or something?

inexorabletash:
Thank you *very* much @asutherland for your detailed feedback. This is extremely helpful, and I'm very happy to have someone else thinking this through. More please!

> abort() post-patch only checks for a state of transaction/finished before throwing. This creates an awkward semantic state where `commit(); abort();` could be invoked and neither method would throw because abort() is okay with being invoked with the transaction in the "committing" state, but either one of the two results could happen depending on the implementation.

True. For reference, in the Chrome implementation it would be racy - the transaction would commit if there was no pending work to be done so it actually committed before the abort showed up. Work might be pending due to implementation details, e.g. multi-stage commit (e.g. blob copies) or if the transaction hadn't processed requests yet (e.g. the transaction hadn't even started, etc).

I'd be okay with making `abort()` throw if the state was_ committing_. Thoughts?

> Similarly, it seems like "fire a success event" can still trigger an (ambiguous) abort even after commit() is called. It directly invokes the "abort a transaction" algorithm without any guards, but I think a transaction state check of [committing, finished] is appropriate.

Agreed, I just missed that. Refined in 9ba00b5ffa5024110bec10ab7c48653e9abb888d

> commit() should probably mention the impact it will have on "success" events

Great point. Added in ef28100ef28684ab14b72ba9b971ae1f683382fb - note that this appears in the "domintro" (non-normative) block, but suggestions for a better location are welcome. (I may have moved too much text out of the normative method descriptions.)

> transaction/lifetime should also mention the "committing". 

Thanks! Done in ca0f393713b8ad0a0b96a5db34e247622cf3350a

> Alternately, the "finished" state could perhaps instead be "completed" and the original semantics of "finished" meaning "abort a transaction" or "commit a transaction" having been invoked (at least once) could be restored.

I was thinking about that too. But since `"complete"` is used as the event for "successfully committed" (vs. `'"abort"` for "aborted") I didn't want to overload that already confusing term. Alternate ideas?

inexorabletash:
> > I'd be okay with making abort() throw if the state was _committing_. Thoughts?
> This sounds like the right choice to me.

Done in a40328bbd3999ba8e174e7a4218b790c432e53a6 ... but in a way that may not be web-compatible, at least not without further rework of the PR. Today you can call abort() at any point, even when the transaction is _inactive_, and thus e.g. when the auto-commit has already begun. Since the PR flips the transaction into the new _committing_ state, that would now throw. Requiring abort() to be called only when the transaction is _active_ would have been a better design from the beginning to avoid a race, but it's probably too late for that.

I don't want to introduce yet another state, so we'll see how this goes.

asutherland:
I agree that additional constraint on abort() would have been nice.

Spec-wise, I think we're in the clear because, modulo some phrasing ambiguity, I think one can interpret the auto-commit step 7 in [IDB 2.0 2.7.1](https://www.w3.org/TR/IndexedDB/#transaction-lifetime-concept) as saying "run commit a transaction, that could still fail over to run abort a transaction".  And I at least interpreted running either algorithm as marking the transaction finished given how the spec abstracted away the actual backend database work.

Web-wise, it's at least the case that Firefox's auto-commit logic [already sets the transaction's internal state to COMMITTING](https://searchfox.org/mozilla-central/rev/97d488a17a848ce3bebbfc83dc916cf20b88451c/dom/indexedDB/IDBTransaction.cpp#373) which will cause [content calls to abort() to throw](https://searchfox.org/mozilla-central/rev/97d488a17a848ce3bebbfc83dc916cf20b88451c/dom/indexedDB/IDBTransaction.cpp#775).  If Chrome has a race at the current time as it relates to abort(), then perhaps you can think of the change as aligning with Firefox's non-racing implementation ;).

index.bs
```diff
@@ -1018,78 +1045,83 @@ The <dfn>lifetime</dfn> of a
...
-    results must be returned in the order the requests were placed
-    against a specific transaction. There is no guarantee about the
-    order that results from requests in different transactions are
-    returned. Similarly, the transaction [=transaction/modes=] ensure that two
-    requests placed against different transactions can execute in any
-    order without affecting what resulting data is stored in the
-    database.
+    When a transaction is created its [=transaction/state=] is initially [=transaction/active=].
+
+1. When an implementation is able to enforce the constraints defined for the transaction [=transaction/scope=] and [=transaction/mode=], defined [below](#transaction-scheduling], the implementation must [=queue a task=] to <dfn lt="start|started">start</dfn> the transaction asynchronously.
```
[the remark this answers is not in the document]
inexorabletash: Tab has been advocating for line breaking at logical (sentence, clause) boundaries rather than to just wrap at 80 columns and reflowing, since that makes future diffs easier to parse. That said, there are some places this could break.

andreas-butler:
Hi all, chrome is getting ready to move forward with implementation of IDBTransaction.comit(). As such any new or evolved thoughts/input regarding the proposed spec changes would be much appreciated at the time, otherwise our intention is for our implementation to comply with the details of this PR.

@asutherland @beidson @aliams @nolanlawson

asutherland:
Can we mark the new commit() method as `[SecureContext]`?  I was just reviewing our (Mozilla's) [exposure guidelines](https://wiki.mozilla.org/ExposureGuidelines) and I was reminded of our [Secure Contexts Everywhere](https://blog.mozilla.org/security/2018/01/15/secure-contexts-everywhere/) policy.

aliams:
@asutherland are you concerned about potential abuse of this method?

asutherland:
@aliams No, Mozilla just has a bias towards requiring SecureContext on new stuff.  I think the blog post and linked TAG github issue covers it.  And in the case of something like 'commit' which currently would stand in isolation and need to be feature-detected, in theory Firefox could ship it SecureContext even if no one else did.

aliams:
Based on my reading of the revisions, it sounds like calling commit would actually flush changes to disk. I  think that there's a concern with allowing malicious websites to flush to disk as they please because they'd be able to cause hardware wear among other things.

pwnall:
@aliams My understanding of this feature is that it lets sites hint to the implementation that they're done with the transaction. It's intended to enable optimizations, not to be a new capability.

A malicious site can still cause a lot of disk writes by opening up many transactions and letting them auto-commit. Committing is still asynchronous, so the implementation has the opportunity to rate-limit and/or batch transactions.

Does this help reason about abuse potential for the new API?

aliams:
The spec currently states (under 2.7.1):
> That is, either all of the changes must be written, or if an error occurs, such as a disk write error, the implementation must not write any of the changes to the database.

This would seem to indicate that we'd need to actually flush to disk?

aliams:
Reading more into this, it seems that @pwnall's statements make sense. Here's the relevant part:
> An explicit call to commit() will _initiate_ a commit

(emphasis on initiate)

pwnall:
@aliams Thank you for explaining!

I read that paragraph as stating that the transaction must be committed in an atomic write. I understood the sentence you quoted as a brief explanation what "atomic" means. I think the intent is only to state that the IDB API user should not be able to observe any partial writes.

index.bs
```diff
@@ -1243,7 +1279,13 @@ operation.
 
 <div dfn-for=request>
 
+A [=/request=] has a <dfn>processed flag</dfn> which is initially unset.
+This flag is set when the operation associated with the request has been executed.
+
+A [=/request=] is said to be <dfn>processed</dfn> when its [=request/processed flag=] is set.
+
 A [=/request=] has a <dfn>done flag</dfn> which is initially unset.
+This flag is set when the request of the operation associated with the request are available.
```
dmurph: This line doesn't make sense to me

index.bs
```diff
@@ -2282,7 +2324,8 @@ when invoked, must run these steps:
             1. Set |request|'s [=request/error=] to |result|.
             1. Set |request|'s [=request/done flag=].
             1. [=Fire an event=] named <a event>`error`</a> at |request| with its
-                {{Event/bubbles}} and {{Event/cancelable}} attributes initialized to true.
+                {{Event/bubbles}} and {{Event/cancelable}} attributes
```
dmurph: Why did we add this line break?

index.bs
inexorabletash: Oops, should say "This flag is set when the results of the operation...."

index.bs
inexorabletash: No idea. Might have been my text editor.

andreas-butler:
Commenting here for the record:

@inexorabletash, @dmurph, and I discussed how to handle requests with errors on them in the context of explicit commit. The primary question dealt with the cases in which an explicitly committed transaction with errored requests should be aborted versus allowed to commit. 

The upshot was an agreement that if, at the time the backend receives a commit signal, the backend is certain that any and all errors on requests belonging to a transaction have been handled in onerror events on the frontend, then that transaction can be committed, otherwise it should be aborted. This is summarized in the attached diagrams.

explicit_commit_error_abort

explicit_commit_error_commit

explicit_commit_error_maybe

pwnall:
@andreas-butler Kudos for the great diagrams!

index.bs
```diff
@@ -5188,14 +5249,43 @@ when invoked, must run these steps:
...
 must run these steps:
 
-1. If this [=/transaction=] is [=transaction/finished=], [=throw=]
-    an "{{InvalidStateError}}" {{DOMException}}.
+1. If this [=/transaction=]'s [=transaction/state=] is [=transaction/committing=]
+    or [=transaction/finished=],
+    then [=throw=] an "{{InvalidStateError}}" {{DOMException}}.
 
-1. Unset the [=/transaction=]'s [=transaction/active flag=] and run the
+1. Set the [=/transaction=]'s [=transaction/state=] to [=transaction/active=] and run the
```
brettz9: Shouldn't this say to set the transaction state to *inactive*?

inexorabletash: Yes, nice catch.

PR #269, opened 2019-06-05
Factor common steps out of add()/put(). 

### LINKED ISSUE #261: Extract common steps of `IDBObjectStore`'s `put()` and`add()`
Currently the 12 steps are almost identical, the only difference being (as far as I can tell) the value of the *no-overwrite* flag passed to [store a record into an object store](https://w3c.github.io/IndexedDB/#store-a-record-into-an-object-store). Extracting them into a separate named procedure would make it easier for users and implementers to understand the difference between the two methods.

inexorabletash:
Editorial only. Resolves #261

The steps between add() and put() operations on IDBObjectStore are identical except for the value of the no-overwrite flag. Factor out the common steps into a named procedure to save space and improve readability.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/269.html" title="Last updated on Jun 6, 2019, 11:08 PM UTC (1933437)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/269/90d61ff...1933437.html" title="Last updated on Jun 6, 2019, 11:08 PM UTC (1933437)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3035,15 +3035,26 @@ and false otherwise.
   </dl>
 </div>
 
-<div class=algorithm>
 
 The <dfn method for=IDBObjectStore>put(|value|, |key|)</dfn> method,
-when invoked, must run these steps:
+when invoked, must return the result of running the steps to [=add or
+put=] with this [=/object store handle=], |value|, |key| and the
+|no-overwrite flag| unset.
 
-1. Let |transaction| be this [=/object store handle=]'s
+The <dfn method for=IDBObjectStore>add(|value|, |key|)</dfn> method,
+when invoked, must return the result of running the steps to [=add or
+put=] with this [=/object store handle=], |value|, |key| and the
+|no-overwrite flag| set.
+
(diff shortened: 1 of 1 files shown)


inexorabletash:
@ytausky @aliams - can you take a look?

index.bs
```diff
@@ -3035,15 +3035,26 @@ and false otherwise.
...
+<div class=algorithm>
+
+The steps to <dfn>add or put</dfn> are as follows. The algorithm
+takes four arguments: |handle|, |value|, |key|, and |no-overwrite flag|.
+
+1. Let |transaction| be |handle|'s
     [=object-store-handle/transaction=].
 
-1. Let |store| be this [=/object store handle=]'s
+1. Let |store| be this |handle|'s's
```
brettz9: Extra `'s`

inexorabletash: Oops, good catch!

PR #272, opened 2019-06-06
Editorial: Correct linking for IDBKeyRange lowerBound/upperBound

### LINKED ISSUE #271: Definitions for IDBKeyRange lowerBound and upperBound aren't linked correctly.
These methods in the IDL block don't link correctly to the algorithms defining them. Probably because IDL defines `lowerBound(lower, open)` while the algorithm defines `lowerBound(lower, lowerOpen)`.

inexorabletash:
Definitions and IDL weren't linked due to different argument
names. Fix it, and unify the prose a bit. Resolves #271

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/272.html" title="Last updated on Jun 7, 2019, 12:28 AM UTC (fc23c72)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/272/d401d30...fc23c72.html" title="Last updated on Jun 7, 2019, 12:28 AM UTC (fc23c72)">Diff</a>


inexorabletash:
@mkruisselbrink - can you review?

Thanks!

mkruisselbrink:
The linking looks good, the prose seems a bit weird, as now you're mixing "setting a flag" with "setting a flag to true/false". At least the old prose was consistent with a flag being either set or not set.

I don't remember what the latest thoughts are on flags vs booleans but we should at least try to be consistent.

inexorabletash:
The old text was inconsistent between the two algorithms:

> Create and return a new key range with lower bound set to lowerKey, **lower open flag set if lowerOpen is true**, upper bound set to null and upper open flag set.

> Create and return a new key range with lower bound set to null, lower open flag set, upper bound set if upperKey, and **upper open flag set to upperOpen**.

I picked the latter example to standardize on, but can switch to the former.

(Latest thing is booleans, not flags, but that's a bigger change.)

inexorabletash:
Also, noticed there's a "upper bound set _if_ upperKey" typo that should be "upper bound set _to_ upperKey"

PR #275, opened 2019-06-11
Use Infra terms to simplify text.

inexorabletash:
* Define array keys as "byte sequence" and compare with "byte less than"
* Compare strings using "code unit less than"
* Make a few lists/sets the Infra types, and use items, contains, append, and so on.
* Simplify key comparison steps for non-equal types.
* Linkify "Assert:"

No normative changes.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/275.html" title="Last updated on Jun 12, 2019, 9:42 PM UTC (8499394)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/275/ce86e18...8499394.html" title="Last updated on Jun 12, 2019, 9:42 PM UTC (8499394)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -695,9 +695,7 @@ To <dfn>compare two keys</dfn> |a| and |b|, run these steps:
         1. Let |i| be 0.
         1. While |i| is less than |length|, then:
 
-            1. Let |u| be the [=/key=] at |va|[|i|].
-            1. Let |v| be the [=/key=] in |vb|[|i|].
-            1. Let |c| be the result of recursively running [=compare two keys=] with |u| and |v|.
+            1. Let |c| be the result of recursively running [=compare two keys=] with |va|[|i|] and |vb|[|i|].
             1. If |c| is not 0, return |c|.
             1. Increase |i| by 1.
 
--- a/index.bs
+++ b/index.bs
@@ -695,15 +695,9 @@ To <dfn>compare two keys</dfn> |a| and |b|, run these steps:
         1. Let |i| be 0.
         1. While |i| is less than |length|, then:
 
-<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


inexorabletash:
The behavior is exactly what was present previously. Sorting is identical to string sorting in JavaScript.

aphillips:
Okay, cool. I mean, it's not ideal, but it makes sense.

inexorabletash:
For the record, #38 tracks locale-aware sorting.

aphillips:
@inexorabletash Yes. In this case, though, I was concerned that code point sorting is better than code unit sorting. In any case, your explanation makes sense.

inexorabletash:
@aliams - can you take a look at this one too?

(It'll conflict with the other one, so whichever comes second will need a rebase)

index.bs
```diff
@@ -688,50 +677,32 @@ To <dfn>compare two keys</dfn> |a| and |b|, run these steps:
...
       <dt>*array*</dt>
       <dd>
-        1. Let |length| be the lesser of |va|'s length and |vb|'s length.
+        1. Let |length| be the lesser of |va|'s [=list/size=] and |vb|'s [=list/size=].
         1. Let |i| be 0.
         1. While |i| is less than |length|, then:
 
-            1. Let |u| be the [=/key=] in |va| at index |i|.
-            1. Let |v| be the [=/key=] in |vb| at index |i|.
+            1. Let |u| be the [=/key=] at |va|[|i|].
```
aliams: At or in?

inexorabletash: I was thinking the temporary _u_ and _v_ variables could be removed entirely, which makes this moot. I'll do that.

PR #279, opened 2019-06-12
Editorial: Replace flag set/unset with true/false.

inexorabletash:
Rather than "setting" and "unsetting" flags, treat flag properties
as booleans (either true or false).

Replace:
 * "Set foo flag" with "Set foo flag to true"
 * "Unset foo flag" with "Set foo flag to false"
 * "If foo is set, ..." with "If foo is true, ..."
 * "If foo is unset, ..." with "If foo is false, ..."

Also clarify a few places that refer to "the foo flag"
to be more explicitly "bar's foo flag".

No normative behavior changes.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -464,7 +464,7 @@ the [=/connection=] is closed the
 [=connection/version=] does not change.
 
 Each connection has a <dfn>close pending flag</dfn> which is initially
-unset.
+false.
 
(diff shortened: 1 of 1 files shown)


inexorabletash:
Another one, @aliams 

Not urgent, obviously.

PR #301, opened 2019-09-26
Sketch out spec text for transaction durability option

### LINKED ISSUE #50: Define flush-to-disk guarantees and control
Define flushing guarantees and expose control
e.g. make transactions not flush-to-disk by default, but introduce a "readwriteflush" mode or {flush: true} option

Currently, FF does not flush (but wants an option), Chrome always flushes.

inexorabletash:
Closes #50

 * [x] Modified Web platform tests ([tentative tests](https://github.com/web-platform-tests/wpt/blob/master/IndexedDB/transaction-relaxed-durability.tentative.any.js))

Implementation commitment:

 * [ ] WebKit (???)
 * [x] Chromium (https://www.chromestatus.com/feature/5730701489995776)
 * [ ] Gecko (???)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/301.html" title="Last updated on Nov 26, 2019, 7:37 PM UTC (27681a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/301/95d8712...27681a2.html" title="Last updated on Nov 26, 2019, 7:37 PM UTC (27681a2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -939,6 +939,21 @@ following:
     instead is created automatically when an
     <a event>`upgradeneeded`</a> event is fired.
 
+A [=/transaction=] has a <dfn>durability hint</dfn>. This is a hint to the user agent of whether to prioritize performance or durability when committing the transaction. The [=transaction/durability hint=] is one of the following:
+
+: {{"strict"}}
+:: The user agent may consider that the [=/transaction=] has successfully [=transaction/committed=] only after verifying that all outstanding changes have been successfully written to a persistent storage medium.
+: {{"relaxed"}}
+:: The user agent may consider that the [=/transaction=] has successfully [=transaction/committed=] as soon as all outstanding changes have been written to the operating system, without subsequent verification.
+: {{"default"}}
+:: The user agent should use its default durability behavior for the storage [=/bucket=]. This is the default for [=/transactions=] if not otherwise specified.
+
+<aside class=note>
+  In a typical implementation, {{"strict"}} is a hint to the user agent to flush any operating system I/O buffers before a {{"complete"}} event is fired. While this provides greater confidence that the changes will be persisted in case of subsequent operating system crash or power loss, flushing buffers can take significant time and consume battery life on portable devices.
+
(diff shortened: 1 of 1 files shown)


pwnall:
I don't think we discussed whether we'd want options to be an alternative to mode. Worth figuring this out before we commit to the API shape described here?

asutherland:
Excellent phrasing on the spec text.  I don't think I personally have an opinion on the signature other than the overload stuff at https://heycam.github.io/webidl/#idl-overloading is complicated and the presence of optional arguments complicates things.  (Specifically, in our attempt to eliminate Firefox's overload of `open` and use-counting we've run into things.)  It seems reasonable to me to stick to not having any overloads if we think we can avoid future overloads, which would be an argument for the current pull request's approach.  If we'll want overloads in the future, we maybe should consider all of this more strongly.

inexorabletash:
> options to be an alternative to mode

As a third argument, it's a graceful fallback for older browsers - silently ignored. Which is great IMHO since it's just a hint. And it avoids an overload, which we're generally allergic to. Open to counter-arguments, though.

A few things the patch needs:

* It needs to say "The durability hint is one of the following:"
* It needs to explicitly say that the default for transaction objects is "default" (e.g. for versionchange transactions)
* The text should not just be tucked in to the bottom of the transaction definition - it might make more sense a few paragraphs up.
* Do we want to throw if a non-default option is set for readonly transactions? IMHO: no.

pwnall:
Do we want to throw if a non-default option is set for readonly transactions? -- I agree with you.

It makes more sense to me that all transactions have a durability property that behaves in the same way. Since it's intended to be a hint, the hint will be ignored in read-only transactions.

inexorabletash:
Any objections to going forward with merging this?

asutherland:
I tried to be cool in December and use a thumbs-up emoji, but I don't think that generated a notification, so to clarify, I'm good with merging this.

inexorabletash:
Hah! Thanks for remembering.

PR #311, opened 2019-11-20
Disallow proxies during key conversion

### LINKED ISSUE #309: Array proxies as keys?
I noticed that this throws in both Chrome and Firefox:

```js
const k = [1,2,3];
const p = new Proxy(k, {});
indexedDB.cmp(k, p);
```
In Chrome, at least, the proxy fails an internal "is this an array?" check. 

In the spec [convert a value to a key](https://w3c.github.io/IndexedDB/#convert-value-to-key) calls ECMAScript's [IsArray](https://tc39.es/ecma262/#sec-isarray) which would seem to return true for array proxies. 

At the very least, we don't have tests. 

Should we update the spec to exclude proxies, e.g. replacing _"if IsArray(input)"_ with _"if input is an Array exotic object"_?

Alternately, should array proxies be supported? What are the consequences of allowing these? Would the value&rarr;key conversion algorithm potentially blow up in more places?

inexorabletash:
Both Gecko and Chromium disallow using proxies when specifying keys, e.g.:
```js
const k = [1,2,3];
const p = new Proxy(k, {});
indexedDB.cmp(k, p); // throws
```
Enshrining this seems like a good idea.

Closes #309

 * [x] Modified Web platform tests (https://github.com/web-platform-tests/wpt/pull/20352)

Implementation commitment:

 * [x] WebKit (already has this behavior)
 * [x] Chromium (already has this behavior)
 * [x] Gecko (already has this behavior)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/311.html" title="Last updated on Nov 22, 2019, 8:26 PM UTC (61548f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/311/98099f2...61548f8.html" title="Last updated on Nov 22, 2019, 8:26 PM UTC (61548f8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -38,33 +38,32 @@ spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
             text: !
             text: ?
         text: abrupt completion; url: sec-completion-record-specification-type
-        text: ReturnIfAbrupt; url: sec-returnifabrupt
-        text: IdentifierName; url: prod-IdentifierName
-        text: Type; url: sec-ecmascript-data-types-and-values
-        text: HasOwnProperty; url: sec-hasownproperty
-        text: Get; url: sec-get-o-p
+        text: Array; url: sec-array-objects
+        text: array exotic object; url: array-exotic-objects
+        text: Array.prototype.sort; url: sec-array.prototype.sort
+        text: ArrayBuffer; url: sec-arraybuffer-objects
         text: CreateDataProperty; url: sec-createdataproperty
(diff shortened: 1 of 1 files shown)


inexorabletash:
Test PR uploaded

inexorabletash:
@asutherland - please take a look?

inexorabletash:
Verified that Safari 13 already exhibits this behavior, so I think we're good.

Issue #291, opened 2019-08-22
How a transaction is implemented as a queue hang


liangzhuang327:
Generally speaking, the above requirements mean that any transaction which has an overlapping scope with a read/write transaction and which was created after that read/write transaction, can’t run in parallel with that read/write transaction.（Mentioned in [W3C](https://w3c.github.io/IndexedDB/#transaction-scheduling)）

And my question is :
How does indexedDB do to suspend the remaining transactions when multiple read/write transactions are opened at the same time? Is it done with the browser's Event Loop?

inexorabletash:
The transaction processing model is somewhat underspecified (and has been the case since the initial IDB spec); the relevant text is in https://w3c.github.io/IndexedDB/#transaction-scheduling and https://w3c.github.io/IndexedDB/#transaction-lifecycle which rather informally defines the preconditions for a transaction to start.

Ideally, the spec would define transaction scheduling e.g. as a [parallel queue](https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism). The steps are not run synchronously with any agent's (document's/worker's) event queue.

PRs that firm up transaction behavior here would be welcome.

PR #319, opened 2020-02-27
Simplify transaction starting constraints to match reality

### LINKED ISSUE #253: Disallow starting readwrite transactions while readonly transactions are running?
The text in [transaction lifetime](https://w3c.github.io/IndexedDB/#transaction-lifetime-concept) about "when a transaction can be started" doesn't define an algorithm (unlike e.g. [web locks](https://wicg.github.io/web-locks/)). Instead, it just defines preconditions. Specifically for this case:

> * ... As long as a read-only transaction is running, the data that the implementation returns through requests created with that transaction must remain constant ...

Which is then followed by the note: 

> There are a number of ways that an implementation can ensure this. The implementation could prevent any read/write transaction, whose scope overlaps the scope of the read-only transaction, from starting until the read-only transaction finishes. Or the implementation could allow the read-only transaction to see a snapshot of the contents of the object stores which is taken when the read-only transaction started.

Chrome implemented snapshots, and therefore allowed a readwrite transaction to start even while a readonly transaction overlapping scope was still running. e.g.

```js
const tx1 = db.transaction('store', 'readonly');
const tx2 = db.transaction('store', 'readwrite');
```

Both tx1 and tx2 could start immediately; operations within tx1 would see a snapshot of the data.

Other engines (Firefox, Safari, Edge) did not implement this; tx2 would not start until tx1 finished.

Chrome is changing this as part of a rework of the implementation; we actually had a non-WPT test asserting this behavior, which broke.

Following this change, all engines will behave the same way - the readonly transactions must finish before the readwrite transaction can start. Should we codify that in the spec and WPTs?

inexorabletash:
All implementations have a strict ordering of transactions with
overlapping scopes; read-only transactions can run in parallel but
block a later read/write transaction from starting, and the read/write
transactions similarly block later read/write and read-only
transactions.

Tighten up the constraints definition to something precise, and move
the wordy implications into a non-normative aside.

Also, define "overlapping scopes" as a term.

Closes #253

 * [ ] Verify existing WPT coverage, or add it

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/319.html" title="Last updated on Mar 2, 2020, 6:51 PM UTC (ef63c28)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/319/093afa4...ef63c28.html" title="Last updated on Mar 2, 2020, 6:51 PM UTC (ef63c28)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6817,6 +6817,7 @@ For the revision history of the second edition, see [that document's Revision Hi
 * Restrict array keys to [=/Array exotic objects=] (i.e. disallow proxies). ([Issue #309](https://github.com/w3c/IndexedDB/issues/309))
 * Transactions are now temporarily made inactive during clone operations.
 * Added {{IDBTransactionOptions/durability}} option and {{IDBTransaction/durability}} attribute. ([Issue #50](https://github.com/w3c/IndexedDB/issues/50))
+* Specified [[#transaction-scheduling]] more precisely and disallow starting read/write transactions while read-only transactions with overlapping scope are running. ([Issue #253](https://github.com/w3c/IndexedDB/issues/253))
 
 <!-- ============================================================ -->
 # Acknowledgements # {#acknowledgements}
--- a/index.bs
+++ b/index.bs
@@ -1114,15 +1114,15 @@ They will return true if any transactions were cleaned up, or false otherwise.
   each [=/transaction=].
 </aside>
 
-<!-- ============================================================ -->
-### Transaction scheduling ### {#transaction-scheduling}
-<!-- ============================================================ -->
-
 An event with type <dfn event>`complete`</dfn> is fired at
 a [=/transaction=] that has successfully [=transaction/committed=].
 
 An event with type <dfn event>`abort`</dfn> is fired at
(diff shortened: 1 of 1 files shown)


inexorabletash:
Re: 1 - agreed. [Web Locks](https://wicg.github.io/web-locks/) in theory has that level of detail. I'd like to get that here at some point. A later PR, though. :)

Re: 2 - that's easy to add.... and great to document; I'd missed that WPT PR.

inexorabletash:
I added normative text:

> Implementations may impose additional constraints. For example, implementations are not required to run non-overlapping read/write transactions in parallel, or may impose limits on the number of running transactions.

The reason being, up in the lifecycle section it says:

> When an implementation is able to enforce the constraints for the transaction’s scope and mode, defined below, the implementation must queue a task to start the transaction asynchronously.

... which seems reasonable as the starting point for a more rigorous algorithm (per #291).

asutherland:
Looks great, thanks!  Your continued evolution of the spec is greatly appreciated!

inexorabletash:
Tests are in. Merging.

PR #330, opened 2020-04-14
Editorial: Use settings object definition.

### LINKED ISSUE #329: Make "global scope" references more concrete
There are uses such as "origin be the origin of the global scope used to access this"

This can be more formally stated as: "origin be the current settings object's origin"

thanks @mkruisselbrink

inexorabletash:
Closes #329

Use the formal definition for environment settings object rather than the imprecise "global scope used to access this".

No normative behavior changes, just being more precise.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/330.html" title="Last updated on Apr 14, 2020, 9:03 PM UTC (656a17f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/330/b226ac6...656a17f.html" title="Last updated on Apr 14, 2020, 9:03 PM UTC (656a17f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2201,7 +2201,7 @@ when invoked, must run these steps:
 
 1. If |version| is 0 (zero), [=throw=] a [=TypeError=].
 
-1. Let |origin| be the [=/current settings object=]'s [=environment settings object/origin=].
+1. Let |origin| be the [=/relevant settings object=]'s [=environment settings object/origin=].
 
 1. If |origin| is an [=opaque origin=], [=throw=] a
     "{{SecurityError}}" {{DOMException}} and abort these steps.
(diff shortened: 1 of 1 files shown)


inexorabletash:
Ah, you're right, relevant is probably correct. 

(Aside: I need to understand why NFS is using a mix of current vs. relevant.)

mkruisselbrink:
Relevant realm/settings object needs to be applied to a particular platform object (i.e. `this`).

(which coincidently is also part of the reason of the mix of current and relevant in NFS, static methods like getSystemDirectory don't have a |this|/platform object, so relevant realm doesn't make sense in that context, hence why that uses the current settings object instead (which also matches what our bindings do for CallWith=ScriptState on static methods). Most of the other cases of current realm in NFS should probably be relevant though, I'll need to fix that...)

PR #334, opened 2020-05-15
Define Indexed DB as a storage endpoint, use hooks

inexorabletash:
WORK IN PROGRESS - not ready to merge

For https://github.com/whatwg/storage/issues/90

Bikeshed hasn't picked up the new terms from Storage yet.

There should be no behavior changes here.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/334.html" title="Last updated on Jun 7, 2021, 5:16 PM UTC (c16bb29)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/334/22382f6...c16bb29.html" title="Last updated on Jun 7, 2021, 5:16 PM UTC (c16bb29)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/334.html" title="Last updated on Apr 11, 2025, 6:49 PM UTC (45bdea5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/334/1b311f3...45bdea5.html" title="Last updated on Apr 11, 2025, 6:49 PM UTC (45bdea5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -43,7 +43,7 @@ spec: ecma-262; urlPrefix: https://tc39.github.io/ecma262/
 spec: storage; urlPrefix: https://storage.spec.whatwg.org/
     type: dfn
         text: storage bucket; url: storage-bucket
-        text: storage key; url: storage-key
+        text: storage identifier; url: storage-identifier
 </pre>
 
 <style>
@@ -396,14 +396,14 @@ To <dfn>create a sorted name list</dfn> from a [=/list=] |names|, run these step
 ## Database ## {#database-construct}
 <!-- ============================================================ -->
 
-Each [=/storage key=] has an associated set of [=/databases=]. A
+Each [=/storage bucket=] has an associated set of [=/databases=]. A
 <dfn>database</dfn> has zero or more [=/object stores=] which
 hold the data stored in the database.
 
 <div dfn-for=database>
 
 A [=/database=] has a <dfn>name</dfn> which identifies it within a
-specific [=/storage key=]. The name is a [=/name=],
+specific [=/storage bucket=]. The name is a [=/name=],
 and stays constant for the lifetime of the database.
 
 A [=/database=] has a <dfn>version</dfn>. When a database is first
@@ -418,6 +418,17 @@ A [=/database=] has at most one associated <dfn>upgrade transaction</dfn>,
 which is either null or an [=/upgrade transaction=], and is initially null.
 
 </div>
+
+<!-- ============================================================ -->
+### Database storage ### {#database-storage}
+<!-- ============================================================ -->
+
+Indexed DB is a [=/storage endpoint=], with the [=/storage identifier=] `"indexedDB"`.
+
+A <dfn>pumpkin</dfn> is a [=/struct=] consisting of a <dfn for="pumpkin">queue</dfn> (a [=/connection queue=]) and a <dfn for="pumpkin">database</dfn> (a [=/database=]).
+
+A <dfn>new pumpkin</dfn> is a [=/pumpkin=] whose [=pumpkin/queue=] is an empty [=/connection queue=], and whose [=pumpkin/database=] is null.
+
 
 <!-- ============================================================ -->
 ### Database connection ### {#database-connection}
@@ -465,7 +476,7 @@ it hasn't already been.
 
 A [=/connection=] may be closed by a user agent in exceptional
 circumstances, for example due to loss of access to the file system, a
-permission change, or clearing of the [=/storage key=]'s storage. If this occurs
+permission change, or clearing the [=/storage bucket=]. If this occurs
 the user agent must run [=close a database
 connection=] with the [=/connection=] and with the <var ignore>forced flag</var> set to true.
 
@@ -1267,8 +1278,8 @@ An [=request/open request=]'s [=get the parent=] algorithm returns null.
 <!-- ============================================================ -->
 
 [=request/Open requests=] are processed in a <dfn>connection queue</dfn>.
-The queue contains all [=request/open requests=] associated with an
-[=/storage key=] and a [=database/name=]. Requests added to the
+The queue contains all [=request/open requests=] associated with
+a [=database/name=] in a [=/storage bucket=]. Requests added to the
 [=/connection queue=] processed in order and each request must run
 to completion before the next request is processed. An open request
 may be blocked on other [=/connections=], requiring those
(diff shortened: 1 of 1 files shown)


annevk:
Looking at this made me notice the "connection queue", which should probably use the storage key as well rather than the origin? Is this a primitive that should move to the Storage Standard?

Also, part of the idea was that you'd no longer need "If origin is an opaque origin" as the Storage Standard would take care of that (and return failure as appropriate).

inexorabletash:
> Looking at this made me notice the "connection queue", which should probably use the storage key as well rather than the origin? Is this a primitive that should move to the Storage Standard?

Agreed it should be decoupled from origin (that's a good mental filter to use in general, thanks). Is it a generic enough concept to move to Storage? Other thoughts are to monkey-patch it on to storage bottle, or make it part of the bottle's contents (e.g. the bottle could have a `"connection queue"` key and a `"databases"` key, where the latter's value a map of the actual databases).

> Also, part of the idea was that you'd no longer need "If origin is an opaque origin" as the Storage Standard would take care of that (and return failure as appropriate).

Agreed, I'll roll that in. Since that error is synchronous in IDB (well, in 2/3 cases), I'll have to rework the algorithms a bit but I think it's fine.

inexorabletash:
Hmmm, actually there is a connection queue per name. So the bottle's map values can be a pair of (queue, database)....? Still thinking this through.

inexorabletash:
And also note that we have no idea what should happen to pending open/delete requests if storage was swapped out. Are they associated with the previous storage? (so the queue is part of storage itself) Or do they apply to whatever the current storage is when they run? Brain hurts...

asutherland:
> And also note that we have no idea what should happen to pending open/delete requests if storage was swapped out. Are they associated with the previous storage? (so the queue is part of storage itself) Or do they apply to whatever the current storage is when they run? Brain hurts...

I think it makes sense for them to be associated with the previous storage.

Step 2 of the proposed replace algorithm at https://github.com/whatwg/storage/issues/18#issuecomment-614751336 is a task that runs on the given agent.  It makes sense that the execution of this task would constitute the start of the new storage epoch, if you will.  Requests made in the before times would be irrelevant.

inexorabletash:
Cool. A few thoughts on how to structure the bottle map...

* _name_ &rarr; (_queue_, _database_)
* _name_ &rarr; _database_, and make _queue_ a property of _database_, and make databases never be deleted, only cleared/reset.
* ("queue", _name_) &rarr; _queue_ and ("database", _name_) &rarr; _database_

Pros and cons for each. Preferences? Other ideas?

asutherland:
>     * _name_ → (_queue_, _database_)

I like this one because:
- Having the key be the name seems like a big win over the compound key approach of `("queue", name)`.
- Separating the queue from the database seems like a nice separation of concerns, especially since the database is something that definitely involves touching disk and is conceptually subject to corruption.  Whereas the connection queue is strictly a runtime concept.

### Blobs / Files

A related question is how IndexedDB-minted Blobs and Files will handle the replace operation and whether this impacts the map.  Gecko definitely invalidates IndexedDB-minted Blobs and Files when Clear-Site-Data and privacy data-clearing operations occur.  From other discussions in the past I have the impression this is also the case in Blink.

The File API Spec doesn't really get into this in [the section on deserialization and the get stream algorithm](https://w3c.github.io/FileAPI/#blob-section) which implies a simplified model where no effort is made to de-duplicate Blob contents or store them to disk, but does leave implementation a broad latitude to [throw errors](https://w3c.github.io/FileAPI/#dfn-error-codes) when get stream is invoked to compensate for the underlying realities.

It seems like we might want to formalize the realities of Blobs/Files now since Clear-Site-Data makes this previous edge-case something content can explicitly trigger instead of a user-initiated edge-case, plus multiple storage buckets presumably would also want to be able to dispose of the underlying blobs and their quota usage in a deterministic fashion.

Doing this might involve a hook where [get stream](https://w3c.github.io/FileAPI/#blob-get-stream) could end up needing to involve some part of the storage hierarchy, in which case it's possible the map might need to store additional data to support this.

mkruisselbrink:
> Blobs / Files

FWIW, I'm working on clarifying this part of the FileAPI spec, with my current thinking being to let others (i.e. IndexedDB, other APIs that produce blobs) define a get stream hook, and then defining all the operations on blobs in terms of that. Unfortunately haven't had as much time to work on that as I would have liked, but it is among the higher priority of spec things I'm working on, also to better define how things work for the Native File System API.

So yes, in that model it would be totally up to IndexedDB to define when/how these blobs get invalidated.

inexorabletash:
Partial update. I needed a name for the (queue,database) struct that exists in the map. I literally called it **pumpkin** here as a placeholder name because I wasn't feeling inspired. So bikeshed away!

This drops the need for most of the imports from Storage, although these are retained:
* **storage bucket** - which exists in the published version, so not new in this PR; this is used to reference durability
* **storage identifier** - this is informative, used in the phrase _Indexed DB is a storage endpoint, with the storage identifier "indexedDB"._ which isn't really necessary but looks nice? We can remove, since it's a registered end point and we don't need to define it. Thoughts?

Most of the remaining references to "origin" end up being fairly illustrative rather than normative definitions. We could probably scrub most of them e.g. "if the origin’s storage is cleared" &rarr; "if the storage is cleared".

index.bs
```diff
@@ -419,6 +419,17 @@ which is either null or an [=/upgrade transaction=], and is initially null.
 
 </div>
 
+<!-- ============================================================ -->
+### Database storage ### {#database-storage}
+<!-- ============================================================ -->
+
+Indexed DB is a [=/storage endpoint=], with the [=/storage identifier=] `"indexedDB"`.
+
+A <dfn>pumpkin</dfn> is a [=/struct=] consisting of a <dfn for="pumpkin">queue</dfn> (a [=/connection queue=]) and a <dfn for="pumpkin">database</dfn> (a [=/database=]).
```
gterzian, line 428: Why are you putting the queues in the map? 

Looking at the current spec, which at [database-construct](https://w3c.github.io/IndexedDB/#database-construct) reads: "Each [storage key](https://storage.spec.whatwg.org/#storage-key) has an associated set of [databases](https://w3c.github.io/IndexedDB/#database).", I think what you need is rather for the map to contain just that set of databases. 

But it's not completely clear to me what a https://storage.spec.whatwg.org/#storage-proxy-map exactly is. Looking at how it is used in https://html.spec.whatwg.org/multipage/webstorage.html#concept-storage-map, it looks like the proxy to the actual database. But not sure how well that would fit in the indexeddb spec. 

At the end of the day, the main goal of the map is to be able to compute things like the https://storage.spec.whatwg.org/#storage-usage and to determine whether it is persisted or not, right?

PR #335, opened 2020-05-18
Editorial: Use WebIDL's new definition conventions for method/etc steps

inexorabletash:
No normative behavior changes, just rewording and some fixes:

    * Per https://github.com/heycam/webidl/issues/730
      * The foo() method steps are: ...
      * The bar getter steps are: ...
      * The bar setter steps are: ...
    * Pull some non-normative descriptive text for getters into notes.
    * Somewhat more correct event handler attribute definitions.
    * Fix a glitch with 'abort a transaction' invocations.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/335.html" title="Last updated on May 19, 2020, 7:51 PM UTC (91967c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/335/c745d4b...91967c7.html" title="Last updated on May 19, 2020, 7:51 PM UTC (91967c7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -60,10 +60,9 @@ spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
         text: Type; url: sec-ecmascript-data-types-and-values
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
         text: Uint8Array; url: sec-typedarray-objects
-spec: webidl; urlPrefix: https://heycam.github.io/webidl/
+spec: storage; urlPrefix: https://storage.spec.whatwg.org/
     type: dfn
-        text: sequence<any>; url: idl-sequence
-        text: sequence<DOMString>; url: idl-sequence
+        text: storage bucket; url: storage-bucket
 </pre>
 
 <style>
@@ -942,7 +941,7 @@ A [=/transaction=] has a <dfn>durability hint</dfn>. This is a hint to the user 
 : {{"relaxed"}}
 :: The user agent may consider that the [=/transaction=] has successfully [=transaction/committed=] as soon as all outstanding changes have been written to the operating system, without subsequent verification.
 : {{"default"}}
-:: The user agent should use its default durability behavior for the storage [=/bucket=]. This is the default for [=/transactions=] if not otherwise specified.
+:: The user agent should use its default durability behavior for the [=/storage bucket=]. This is the default for [=/transactions=] if not otherwise specified.
 
 <aside class=note>
   In a typical implementation, {{"strict"}} is a hint to the user agent to flush any operating system I/O buffers before a {{"complete"}} event is fired. While this provides greater confidence that the changes will be persisted in case of subsequent operating system crash or power loss, flushing buffers can take significant time and consume battery life on portable devices.
@@ -2030,7 +2029,7 @@ The <dfn attribute for=IDBRequest>transaction</dfn> getter steps are to
 return [=/this=]'s [=request/transaction=].
 
 <aside class=note>
-    The {{IDBRequest/transaction}} property can be null for certain requests, such as for [=/requests=] returned from {{IDBFactory/open()}}.
+    The {{IDBRequest/transaction}} getter can return null for certain requests, such as for [=/requests=] returned from {{IDBFactory/open()}}.
 </aside>
 
 The <dfn attribute for=IDBRequest>readyState</dfn> getter steps are to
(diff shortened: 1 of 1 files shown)


inexorabletash:
@domenic maybe you could review this one?

index.bs
```diff
@@ -2005,36 +2005,42 @@ enum IDBRequestReadyState {
...
-must return [=/this=]'s [=request/transaction=]. This
-property can be null for certain requests, such as for [=/requests=]
-returned from {{IDBFactory/open()}}.
+The <dfn attribute for=IDBRequest>transaction</dfn> getter steps are to
+return [=/this=]'s [=request/transaction=].
 
-The <dfn attribute for=IDBRequest>readyState</dfn> attribute's getter
-must return {{"pending"}} if [=/this=]'s [=request/done flag=] is false, and
+<aside class=note>
+    The {{IDBRequest/transaction}} property can be null for certain requests, such as for [=/requests=] returned from {{IDBFactory/open()}}.
```
domenic, line 2033: Maybe "The transaction getter can return null"? Both are correct though.

index.bs
```diff
@@ -2078,14 +2081,10 @@ dictionary IDBVersionChangeEventInit : EventInit {
 };
 </xmp>
 
-The <dfn attribute for=IDBVersionChangeEvent>oldVersion</dfn>
-attribute getter returns the previous version of the database.
+The <dfn attribute for=IDBVersionChangeEvent>oldVersion</dfn> getter steps are to return the value it was initialized to. It represents the previous version of the database.
```
domenic, line 2084: Ideally this concept of "value it was initialized to" would be replaced with explicit associated values, but this is an improvement over the previous version.

index.bs
```diff
@@ -2503,8 +2500,8 @@ must return [=/this=]'s [=connection/version=].
         [=/upgrade transaction=].
 </div>
 
-The <dfn attribute for=IDBDatabase>objectStoreNames</dfn> attribute's
-getter must return a {{DOMStringList}} associated with a [=sorted name list=] of the [=object-store/names=] of
+The <dfn attribute for=IDBDatabase>objectStoreNames</dfn> getter steps are to
+return a {{DOMStringList}} associated with a [=sorted name list=] of the [=object-store/names=] of
 the [=/object stores=] in [=/this=]'s [=object store set=].
```
domenic, line 2505: Preexisting issue: it's not clear whether this returns the same DOMStringList each time, or a new DOMStringList. (The former would certainly be better.)

index.bs
```diff
@@ -2838,21 +2827,22 @@ The {{IDBObjectStore/name}} attribute's setter must run these steps:
 </div>
 
 
-The <dfn attribute for=IDBObjectStore>keyPath</dfn> attribute's getter
-must return [=/this=]'s [=object-store-handle/object store=]'s [=object-store/key path=], or
+The <dfn attribute for=IDBObjectStore>keyPath</dfn> getter steps are to
+return [=/this=]'s [=object-store-handle/object store=]'s [=object-store/key path=], or
 null if none. The [=/key path=] is converted as a {{DOMString}} (if a
 string) or a [=sequence&lt;DOMString&gt;=] (if a list of strings), per
```
domenic, line 2833: Preexisting issue: this could be nicer typographically if you used the awkward `<code><a interface>sequence</a>&lt;<a interface>DOMString</a>&gt;</code>`

index.bs
```diff
@@ -3586,8 +3561,7 @@ invoked, must run these steps:
 
 <div class=algorithm>
```
domenic, line 3562: Preexisting issue: `<div class=algorithm>` is, I believe, not processed by Bikeshed in the same ways as `<div algorithm>`. In particular you're not getting the nice user-facing feature where you can click on a `<var>` and all the instances of it are highlighted, or the spec-developer facing feature of checking for variables that only appear once (usually due to typos).

index.bs
inexorabletash: Yeah... I went looking for how other specs deal with event interfaces and they seem to have not gotten the modernization other places have. Leaving this for now.

domenic: Oh, I missed that this was an event interface. Yeah, then this seems good, for sure.

index.bs
inexorabletash: done

index.bs
inexorabletash: Done.

index.bs
inexorabletash: Ah, interesting - thanks! I'll take this on separately. Generated a lot of errors for variables used only once seems to be a problem, likely in normative blocks outside algorithms that should be non-normative. I'll need to work through them.

index.bs
inexorabletash: Fixed - sadly by inserting "new" in each such site, at least per Chrome's impl. I don't see WPT for this, so it might be web compatible to change. It gets weird for IDBTransaction's `objectStoreNames` property since that can change over time. In any case this isn't a method that's ever going to be a performance bottleneck so it's not a big deal.

PR #336, opened 2020-05-19
Editorial: Use Bikeshed's algorithm and var more better

inexorabletash:

Switch from `<div class=algorithm>` to `<div algorithm>` to get better var highlighting/checking. A consequence of this is fixing issues, the biggest one being places where algorithm A calls into B with named parameters, e.g. "with foo as bar". In that case, "bar" needs to be called out with `<var ignore>`. 

A few other glitches were noticed along the way, e.g. in domintro sections, some non-normative text was correctly marked as asides, and a few incorrect variables were fixed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/336.html" title="Last updated on May 20, 2020, 11:14 PM UTC (2f546e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/336/ff8346e...2f546e3.html" title="Last updated on May 20, 2020, 11:14 PM UTC (2f546e3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3120,7 +3120,7 @@ The <dfn method for=IDBObjectStore>get(|query|)</dfn> method steps are:
 
 1. Let |range| be the result of running [=convert a value to a key range=] with |query| and true. Rethrow any exceptions.
 
-1. Let |operation| be an algorithm to run [=retrieve a value from an object store=] with [=current Realm=], |store| and |range|.
+1. Let |operation| be an algorithm to run [=retrieve a value from an object store=] with the [=/current Realm=], |store|, and |range|.
 
 1. Return the result (an {{IDBRequest}}) of running [=asynchronously execute a request=] with [=/this=] and |operation|.
 
@@ -3191,7 +3191,7 @@ The <dfn method for=IDBObjectStore>getAll(|query|, |count|)</dfn> method steps a
     [=convert a value to a key range=] with |query|.
     Rethrow any exceptions.
 
-1. Let |operation| be an algorithm to run [=retrieve multiple values from an object store=] with [=current Realm=], |store|, |range|, and |count| if given.
+1. Let |operation| be an algorithm to run [=retrieve multiple values from an object store=] with the [=/current Realm=], |store|, |range|, and |count| if given.
 
 
 1. Return the result (an {{IDBRequest}}) of running [=asynchronously execute a request=] with [=/this=] and |operation|.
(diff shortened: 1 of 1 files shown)


inexorabletash:
@domenic - no good deed goes unpunished - can you take a look? Thanks!

index.bs
```diff
@@ -2985,14 +2986,14 @@ To <dfn>add or put</dfn> with |handle|, |value|, |key|, and |no-overwrite flag|,
 
 1. Return the result (an {{IDBRequest}}) of running
     [=asynchronously execute a request=]
-    with |handle| as |source| and
+    with |handle| as <var ignore>source</var> and
     [=store a record into an object store=] as
-    |operation|, using |store|, the |clone| as |value|, |key|, and
+    <var ignore>operation</var>, using |store|, the |clone| as <var ignore>value</var>, |key|, and
     |no-overwrite flag|.
```
domenic, line 2992: This is pretty confusing, as it mixes named and positional parameters. I'd suggest just using "_store_, _value_, _key_, and _no-overwrite flag_".

index.bs
```diff
@@ -3011,19 +3012,21 @@ The <dfn method for=IDBObjectStore>delete(|query|)</dfn> method steps are:
...
-    |null disallowed flag| true. Rethrow any exceptions.
+    <var ignore>null disallowed flag</var> true. Rethrow any exceptions.
 
 1. Return the result (an {{IDBRequest}}) of running
     [=asynchronously execute a request=]
-    with [=/this=] as |source| and
+    with [=/this=] as <var ignore>source</var> and
     [=delete records from an object store=] as
-    |operation|, using |store| and |range|.
+    <var ignore>operation</var>, using |store| and |range|.
```
domenic, line 3021: How I would phrase this (to avoid the "named parameter passing") is "asynchronously execute a request with this and an algorithm that performs delete records from an object store using store and range". I guess you could even factor it into two steps:

1. Let _operation_ be an algorithm that performs delete records from an object store using _store_ and _range_.
1. Asynchronously execute a request with this and _operation_.

inexorabletash: Good idea; this has always been an awkward part of the spec since it went algorithmic.

index.bs
inexorabletash: Done!

inexorabletash:
One more look, @domenic?

PR #350, opened 2021-03-11
automatic publishing

deniak:
PR adds the GH action to publish to gh-pages


siusin (approved):
Thanks! This will generate a snapshot for every PR.

PR #357, opened 2021-06-03
Allow a transaction's scope to change if the transaction is an upgrade

### LINKED ISSUE #356: Calling tx.objectStore after db.createObjectStore
A developer asked me how to add an index to an existing object store, and the answer is:

```js
const request = indexedDB.open('my-db');

request.onupgradeneeded = (event) => {
  const db = request.result;
  const tx = request.transaction;

  if (event.oldVersion < 1) {
    db.createObjectStore('key-val');
  }
  if (event.oldVersion < 2) {
    const store = tx.objectStore('key-val');
    store.createIndex(/* ... */);
  }
};
```

However, I was initially unsure that `tx.objectStore('key-val')` would work. It works in browsers, but the spec suggests otherwise.

https://w3c.github.io/IndexedDB/#dom-idbtransaction-objectstore - `tx.objectStore` gets stores that are in its scope.

https://w3c.github.io/IndexedDB/#transaction-scope - The scope is a "set of object stores" that is "fixed for the lifetime of that transaction".

To me, this suggests that the following should fail:

```js
db.createObjectStore('key-val');
const store = tx.objectStore('key-val');
```

…because the transaction's scope's set of object stores does not include the "key-val" store (since it doesn't exist at the time the transaction is created), and the scope's set of object stores is immutable, so it cannot be added to the set when the store is created.

It seems like the spec should contain a special kind of scope, where the set of object stores is the [connection's set](https://w3c.github.io/IndexedDB/#connection-object-store-set) (which does mutate). Only upgrade transactions may use this scope.

jakearchibald:
Closes #356

This is a non-normative change.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/357.html" title="Last updated on Jun 3, 2021, 9:43 AM UTC (d41d0f1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/357/92322af...d41d0f1.html" title="Last updated on Jun 3, 2021, 9:43 AM UTC (d41d0f1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -900,7 +900,9 @@ and data mutation operations.
 All transactions are created through a [=/connection=], which is the
 transaction's <dfn>connection</dfn>.
 
-A [=/transaction=] has a <dfn>scope</dfn> which is a [=/set=] of [=/object stores=] that the transaction may interact with. A transaction's scope remains fixed for the lifetime of that transaction.
+A [=/transaction=] has a <dfn>scope</dfn> which is a [=/set=] of [=/object stores=] that the transaction may interact with.
+
+Note: A [=/transaction=]'s [=scope=] remains fixed unless the [=/transaction=] is an [=/upgrade transaction=].
 
 Two [=/transactions=] have <dfn lt="overlap|overlapping scope">overlapping scope</dfn> if any [=/object store=] is in both transactions' [=transaction/scope=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4773,15 +4775,7 @@ enum IDBTransactionMode {
...
-The <dfn attribute for=IDBTransaction>objectStoreNames</dfn> getter steps are:
-
-1. If [=/this=] is an [=/upgrade transaction=],
-    then return a new {{DOMStringList}} associated with a [=sorted name list=] of the [=object-store/names=]
-    of the [=/object stores=] in [=/this=]'s [=/connection=]'s [=object store set=].
-
-1. Otherwise, return a new {{DOMStringList}} associated with a [=sorted name list=] of the
-    [=object-store/names=] of the [=/object stores=] in
-    [=/this=]'s [=transaction/scope=].
+The <dfn attribute for=IDBTransaction>objectStoreNames</dfn> getter steps are to return a new {{DOMStringList}} associated with a [=sorted name list=] of the [=object-store/names=] of the [=/object stores=] in [=/this=]'s [=transaction/scope=].
```
jakearchibald, line 4778: This is simpler now, since an update transaction's scope is already the connection's object store set.

inexorabletash (approved):
Nice, thanks for the fix/simplification.

PR #409, opened 2023-07-28
Disambiguate "running" for transactions

### LINKED ISSUE #408: Transaction lifetime verbiage, "running"
The spec refers to _times when a transaction is running_ or _a running transaction_ in several places e.g. [[1]](https://www.w3.org/TR/IndexedDB/#connection-object-store-set)[[2]](https://www.w3.org/TR/IndexedDB/#object-store-handle-index-set)[[3]](https://www.w3.org/TR/IndexedDB/#transaction-scheduling)[[4]](https://www.w3.org/TR/IndexedDB/#upgrade-transaction-construct)[[5]](https://www.w3.org/TR/IndexedDB/#dom-idbdatabase-transaction). It's not clear to me what it means to be in a running state or if it's possible to have a transaction that is not running. Nothing in the "transaction lifecycle" portion mentions when a transaction is "running". I'm not sure if we need a definition of running or if we need to replace the term with other ones that already carry definitions.

"Running" is also used in the name of an algorithm, namely "Running an upgrade transaction". But some of the times the doc uses "running" refer to non-upgrade transactions, so I don't think the definition is "while the steps to run an upgrade transaction are being executed" is the correct definition.

Does "running" simply mean that the transaction exists and is associated with a database? Does it mean any state in the transaction lifecycle prior to "finished"? In some cases, such as "limits on the number of running transactions", I think it might mean after "started" (which is not itself a state) but before the state is set to finished.

I would also suggest changing "Running an upgrade transaction" so we don't get phrases like "run the steps to run a transaction". Maybe it could be "Upgrading a database"?

inexorabletash:
The phrase "a transaction is running" was not normatively defined and confusing given the complex lifecycle of transactions. Introduce a new definition for "live" which covers from a transactions creation until it is finished.

Also, rename "running an upgrade transaction" since steps are often invoked with "run" and that could get awkward. Simplify to "upgrade a database".

No behavior changes, just normative terminology definitions.

Resolves #408. Thanks to @evanstade the issue and suggestions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/409.html" title="Last updated on Aug 3, 2023, 11:21 PM UTC (b70faa9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/409/9cd66f0...b70faa9.html" title="Last updated on Aug 3, 2023, 11:21 PM UTC (b70faa9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1001,7 +1001,7 @@ Transactions are expected to be short lived. This is encouraged by the
 described below.
 
 <aside class=note>
-  Authors can still cause transactions to stay alive for a long time;
+  Authors can still cause transactions to stay [=transaction/live|alive=] for a long time;
   however, this usage pattern is not advised as it can lead to a poor
   user experience.
 </aside>
--- a/index.bs
+++ b/index.bs
@@ -474,7 +474,7 @@ A [=/connection=] has an <dfn>object store set</dfn>, which is
 initialized to the set of [=/object stores=] in the associated
 [=/database=] when the [=/connection=] is created. The contents of the
 set will remain constant except when an [=/upgrade transaction=] is
-running.
+[=transaction/live=].
 
 A [=/connection=]'s [=get the parent=] algorithm returns
 null.
@@ -553,13 +553,13 @@ An [=/object store handle=] has an <dfn>index set</dfn>, which is
 initialized to the set of [=/indexes=] that reference the associated
 [=object-store-handle/object store=] when the [=/object store handle=]
 is created. The contents of the set will remain constant except when
-an [=/upgrade transaction=] is running.
+an [=/upgrade transaction=] is [=transaction/live=].
 
 An [=/object store handle=] has a <dfn>name</dfn>, which is
 initialized to the [=object-store/name=] of the associated
 [=object-store-handle/object store=] when the [=/object store handle=]
 is created. The name will remain constant except when an [=/upgrade
-transaction=] is running.
+transaction=] is [=transaction/live=].
 
 </div>
 
@@ -866,7 +866,7 @@ particular [=/index=] within a [=/transaction=].
 An [=index handle=] has a <dfn>name</dfn>, which is initialized to the
 [=index/name=] of the associated [=index-handle/index=] when the
 [=index handle=] is created. The name will remain constant except when
-an [=/upgrade transaction=] is running.
+an [=/upgrade transaction=] is [=transaction/live=].
 
 </div>
 
@@ -905,7 +905,7 @@ following:
 ::
     The transaction is only allowed to read data. No modifications can
     be done by this type of transaction. This has the advantage that
-    several [=transaction/read-only transactions=] can run at the same time even
+    several [=transaction/read-only transactions=] can be [=transaction/started=] at the same time even
     if their [=transaction/scopes=] are [=transaction/overlapping=], i.e. if they are using the
     same object stores. This type of transaction can be created any
     time once a database has been opened.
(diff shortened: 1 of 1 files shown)


inexorabletash:
@evanstade could you take a look?

index.bs
```diff
@@ -5297,12 +5299,12 @@ created [=/request=] belongs to is [=transaction/aborted=] using the steps to
...
 <!-- ============================================================ -->
-## Running an upgrade transaction ## {#upgrade-transaction-steps}
+## Upgrading a database ## {#upgrade-transaction-steps}
 <!-- ============================================================ -->
 
 <div algorithm>
 
-To <dfn>run an upgrade transaction</dfn> with |connection| (a [=/connection=])
+To <dfn>upgrade a database</dfn> with |connection| (a [=/connection=])
 which is used to update the [=/database=], a new |version| to be set
```
evanstade, line 5308: > which is used to update the database

This part seems a little superfluous maybe.

inexorabletash: Thanks. Simplified a bunch, now that "database" is clearly the object of the phrase.

index.bs
```diff
@@ -1171,19 +1173,19 @@ handler.
...
   The {{IDBOpenDBRequest/upgradeneeded!!event}} event isn't fired, and thus the
   [=/upgrade transaction=] isn't started, until all other
   [=/connections=] to the same [=/database=] are closed. This ensures
   that all previous transactions are [=transaction/finished=]. As long
-  as an [=/upgrade transaction=] is running, attempts to open more
+  as an [=/upgrade transaction=] is [=transaction/live=], attempts to open more
   [=/connections=] to the same [=/database=] are delayed, and any
   attempts to use the same [=/connection=] to start additional
   transactions by calling {{IDBDatabase/transaction()}} will throw an
   exception. Thus [=/upgrade transactions=] not only ensure that no
```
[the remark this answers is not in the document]
inexorabletash: Whee, this text basically goes back to 2012. At least it's non-normative now. :) I rephrased a bit, I think the new form is clearer? As it is non-normative, I think the whole sentence could easily be dropped. But presumably it was added to make something explicit to a reader at some point. WDYT?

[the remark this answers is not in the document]
evanstade: I agree that dropping it might ultimately be less confusing since it's essentially restating information from elsewhere in the doc, but less clearly (imo). The new form is better too.

evanstade (approved):

index.bs
```diff
@@ -5297,14 +5301,12 @@ created [=/request=] belongs to is [=transaction/aborted=] using the steps to
...
-## Running an upgrade transaction ## {#upgrade-transaction-steps}
+## Upgrading a database ## {#upgrade-transaction-steps}
 <!-- ============================================================ -->
 
 <div algorithm>
 
-To <dfn>run an upgrade transaction</dfn> with |connection| (a [=/connection=])
-which is used to update the [=/database=], a new |version| to be set
-for the [=/database=], and a |request|, run these steps:
+To <dfn>upgrade a database</dfn> with |connection| (a [=/connection=]), a new |version|, and a |request|, run these steps:
```
a-sully, line 5309: Do you think it's worthwhile to preserve the old ID? There probably don't exist many references to https://www.w3.org/TR/IndexedDB/#run-an-upgrade-transaction out on the interwebs, but if there are they'd be broken

a-sully (approved):
LGTM % optional nits

index.bs
inexorabletash: Good question. After a bit of searching I don't see many references to the old phrase out on the webs, so I'm suspecting the number of direct links into the algorithm from anywhere other than issues in this repo are low. So... leaning to "no" but we can restore the id if anyone complains?

PR #430, opened 2024-10-08
Specify the error code for failures when reading a value from disk

### LINKED ISSUE #423: Specify the exception for read failures arising due to partial data loss on the file system
IndexedDB data stored on the file system can be lost due to factors beyond the browser's control, such as when users directly delete the files containing the data.
Total data loss of the `database` or `object store` is visible to the script as an _absence_ of data, and can be appropriately handled by it.
However, since the current specification does not mandate consistent data storage, there is a chance of partial data loss where the entry for a `record` exists in the `object store` but the contents of the `record` are lost.
For example, both Chromium and Firefox store blobs and large values in files separate from the main database, which only contains references to these files.
When those files are deleted, the references in the database become dangling and attempts to read the data fail permanently.

The specification does not cover the above scenario, and hence the `DOMException` thrown varies by browser:
* Chromium:
  * `name`: `NotFoundError` ([updated recently](https://chromestatus.com/feature/5140210640486400))
  * `message`: "Failed to read large IndexedDB value"
* Firefox:
  * `name`: `UnknownError`
  * `message`: "The operation failed for reasons unrelated to the database itself and not covered by any other error code."

Specifying a common `DOMException` for such cases of partial data loss will help developers handle these scenarios consistently across browsers.

abhishek-shanthkumar:
This PR:
- Updates the description of `UnknownError` to convey that it should be used for transient reasons, per the [Web IDL Standard](https://webidl.spec.whatwg.org/#unknownerror).
- Adds `NotReadableError` from the [Web IDL Standard](https://webidl.spec.whatwg.org/#notreadableerror) to the list of possible exceptions.
- Specifies `NotReadableError` to be returned when reading the value from the underlying storage fails in the `object-store-retrieval-operation` algorithms.

Closes #423

 * [ ] ~Modified Web platform tests (link to pull request)~ A WPT for this scenario does not seem feasible because the location of IndexedDB data on disk differs across browsers and platforms.
 

Implementation commitment:

 * [x] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=362123231)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/430.html" title="Last updated on Oct 14, 2024, 9:34 AM UTC (1489231)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/430/010ecb2...1489231.html" title="Last updated on Oct 14, 2024, 9:34 AM UTC (1489231)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1869,8 +1869,8 @@ usage.
   <tr>
     <td>{{NotReadableError}}</td>
     <td>
-      The operation failed because the file(s) containing the requested
-      data could not be read.
+      The operation failed because the underlying storage containing
+      the requested data could not be read.
     </td>
   </tr>
   <tr>
@@ -1900,7 +1900,7 @@ usage.
     <td>{{UnknownError}}</td>
     <td>
       The operation failed for transient reasons unrelated to the
-      database itself and not covered by any other error.
+      database itself or not covered by any other error.
     </td>
   </tr>
   <tr>
@@ -5643,6 +5643,9 @@ To <dfn>store a record into an object store</dfn> with
 To <dfn>retrieve a value from an object store</dfn> with
 |targetRealm|, |store| and |range|, run these steps:
 
+NOTE:
+    May return a {{DOMException}} if an error occurs while running these steps.
+
 1. Let |record| be the first [=object-store/record=] in |store|'s
     [=object-store/list of records=] whose [=/key=] is [=in=] |range|, if
     any.
@@ -5661,6 +5664,9 @@ To <dfn>retrieve a value from an object store</dfn> with
 
 To <dfn>retrieve multiple values from an object
 store</dfn> with |targetRealm|, |store|, |range| and optional |count|, run these steps:
+
+NOTE:
+    May return a {{DOMException}} if an error occurs while running these steps.
 
 1. If |count| is not given or is 0 (zero), let |count| be infinity.
 
--- a/index.bs
+++ b/index.bs
@@ -1867,6 +1867,13 @@ usage.
     </td>
   </tr>
   <tr>
+    <td>{{NotReadableError}}</td>
+    <td>
+      The operation failed because the file(s) containing the requested
+      data could not be read.
+    </td>
+  </tr>
+  <tr>
     <td>{{QuotaExceededError}}</td>
     <td>
       The operation failed because there was not enough remaining
@@ -1892,8 +1899,8 @@ usage.
   <tr>
     <td>{{UnknownError}}</td>
     <td>
-      The operation failed for reasons unrelated to the database
-      itself and not covered by any other errors.
+      The operation failed for transient reasons unrelated to the
+      database itself and not covered by any other error.
     </td>
   </tr>
   <tr>
@@ -5642,7 +5649,8 @@ To <dfn>retrieve a value from an object store</dfn> with
 
 1. If |record| was not found, return undefined.
 
-1. Let |serialized| be of |record|'s [=/value=].
+1. Let |serialized| be |record|'s [=/value=]. If an error occurs while reading the value from the
+    underlying storage, return a newly [=exception/created=] "{{NotReadableError}}" {{DOMException}}.
 
 1. Return [=ECMAScript/!=] [$StructuredDeserialize$](|serialized|, |targetRealm|).
 
@@ -5664,7 +5672,8 @@ store</dfn> with |targetRealm|, |store|, |range| and optional |count|, run these
 
 1. [=list/For each=] |record| of |records|:
 
-    1. Let |serialized| be |record|'s [=/value=].
+    1. Let |serialized| be |record|'s [=/value=]. If an error occurs while reading the value from the
+        underlying storage, return a newly [=exception/created=] "{{NotReadableError}}" {{DOMException}}.
     1. Let |entry| be [=ECMAScript/!=] [$StructuredDeserialize$](|serialized|, |targetRealm|).
     1. Append |entry| to |list|.
 
--- a/index.bs
+++ b/index.bs
@@ -6717,6 +6717,7 @@ For the revision history of the second edition, see [that document's Revision Hi
 * Added <a href="#accessibility">Accessibility considerations</a> section. ([Issue #327](https://github.com/w3c/IndexedDB/issues/327))
 * Used [[infra]]'s list sorting definition. ([Issue #346](https://github.com/w3c/IndexedDB/issues/346))
 * Added a definition for [=transaction/live=] transactions, and renamed "run an upgrade transaction" to [=/upgrade a database=], to disambiguate "running". ([Issue #408](https://github.com/w3c/IndexedDB/issues/408))
+* Specified the {{DOMException}} type for failures when reading a value from the underlying storage in [[#object-store-retrieval-operation]]. ([Issue #423](https://github.com/w3c/IndexedDB/issues/423))
 
 <!-- ============================================================ -->
 # Acknowledgements # {#acknowledgements}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5642,7 +5649,8 @@ To <dfn>retrieve a value from an object store</dfn> with
 
 1. If |record| was not found, return undefined.
 
-1. Let |serialized| be of |record|'s [=/value=].
+1. Let |serialized| be |record|'s [=/value=]. If an error occurs while reading the value from the
```
inexorabletash, line 5652: Can you update the prose introducing the algorithm to indicate that it may return an error (DOMException) ?

Reviewing the rest of the spec, https://w3c.github.io/IndexedDB/#asynchronously-execute-a-request does have prose like _"if result is an error ..."_ which should process this failure correctly. I note that if we ever allowed storing a DOMException then the behavior would be ambiguous and we'd need to be more explicit here (and elsewhere) about success vs. failure, but that's not currently a problem.

index.bs
```diff
@@ -5664,7 +5672,8 @@ store</dfn> with |targetRealm|, |store|, |range| and optional |count|, run these
 
 1. [=list/For each=] |record| of |records|:
 
-    1. Let |serialized| be |record|'s [=/value=].
+    1. Let |serialized| be |record|'s [=/value=]. If an error occurs while reading the value from the
```
inexorabletash, line 5675: Same as above - can you update the prose introducing the algorithm?

index.bs
```diff
@@ -1866,6 +1866,13 @@ usage.
       not be found.
     </td>
   </tr>
+  <tr>
+    <td>{{NotReadableError}}</td>
+    <td>
+      The operation failed because the file(s) containing the requested
```
inexorabletash, line 1872: Maybe "underlying storage" instead of "file(s)" since we don't require that it actually bottoms out in files?

index.bs
```diff
@@ -1892,8 +1899,8 @@ usage.
   <tr>
     <td>{{UnknownError}}</td>
     <td>
-      The operation failed for reasons unrelated to the database
-      itself and not covered by any other errors.
+      The operation failed for transient reasons unrelated to the
+      database itself and not covered by any other error.
```
inexorabletash, line 1903: In my opinion, this text scopes `UnknownError` a bit too much. If it's any other type of error, a developer should be able to look at the spec and deduce exactly what the cause was; `UnknownError` is basically for anything unspecified.

Maybe "or not covered" instead of "and not covered" ?

inexorabletash (commented):
Overall looks good, a few notes/suggestions.

index.bs
abhishek-shanthkumar: > ... does have prose like "if result is an error ..." which should process this failure correctly.

Yes, I was banking on this prose to complete the picture of how the returned error gets surfaced out of `IDBRequest`, but I agree that explicitly indicating this in the retrieval algorithm would help. I'll update it.

abhishek-shanthkumar: Done

index.bs
abhishek-shanthkumar: Done

index.bs
abhishek-shanthkumar: Done

index.bs
abhishek-shanthkumar: Done

index.bs
```diff
@@ -5636,13 +5643,17 @@ To <dfn>store a record into an object store</dfn> with
 To <dfn>retrieve a value from an object store</dfn> with
 |targetRealm|, |store| and |range|, run these steps:
 
+NOTE:
+    May return a {{DOMException}} if an error occurs while running these steps.
```
inexorabletash, line 5647: Avoid "may" as that has [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) meaning, i.e. it means that the specified behavior is optional for implementations. Also, it should be normative.

https://infra.spec.whatwg.org/#algorithm-declaration has the preferred formulation, which is:

> ..., perform the following steps. They return ...

So maybe:

> ... run these steps. They return a value from the object store, or an error (a DOMException).

index.bs
```diff
@@ -5654,6 +5665,9 @@ To <dfn>retrieve a value from an object store</dfn> with
 To <dfn>retrieve multiple values from an object
 store</dfn> with |targetRealm|, |store|, |range| and optional |count|, run these steps:
 
+NOTE:
```
inexorabletash, line 5668: Same as above.

inexorabletash (changes requested):

index.bs
abhishek-shanthkumar: Thank you for these pointers! I've updated the algorithm prose accordingly. https://infra.spec.whatwg.org/#algorithm-declaration specifies mentioning the _return type_ and not a description of the returned value, so I have used `ECMAScript value` instead and also included `undefined` in the list of possible return types. Please LMK if I've interpreted any of this incorrectly.

index.bs
abhishek-shanthkumar: Updated. I've specified the return type as `sequence<any>` here since that is what we return from the last step of this algorithm. Perhaps
> ... a sequence of ECMAScript values or ...

would be better (and aligned with the return type of the single value retrieval above), but I'm not aware of how implicit the equivalence between `any` and `ECMAScript value` is. Do LMK if there is a common term we can use for both these algorithms.

inexorabletash (changes requested):
Looks great! Thank you for your patience.

Two last non-normative things:

1. Add a note into the [Revision History](https://w3c.github.io/IndexedDB/#revision-history) section, referencing the issue.
2. Add yourself to the [Acknowledgements](https://w3c.github.io/IndexedDB/#acknowledgements) section

(Normally I might do these as a follow-up PR but it's easiest if you do it - thank you!!!)

abhishek-shanthkumar:
Thank you for the review! I've updated these two sections, please take a look.

abhishek-shanthkumar:
Thank you! This wasn't discussed at TPAC.
@asutherland from Gecko [is aligned with](https://github.com/w3c/IndexedDB/issues/423#issuecomment-2375506223) specifying a common error code and discussing further about standardizing handling of DB corruption in general.
No word from WebKit yet. @annevk, could you indicate your position on this PR (and [its issue](https://github.com/w3c/IndexedDB/issues/423))? Happy to learn more about whether this applies to WebKit too!

inexorabletash:
Ah, that's right - inspired by Gecko. I'll go ahead and merge.

index.bs
```diff
@@ -1866,6 +1866,13 @@ usage.
       not be found.
     </td>
   </tr>
+  <tr>
+    <td>{{NotReadableError}}</td>
```
evanstade, line 1870: Might be worth mentioning that this is considered a "permanent" loss or that the data is "irrecoverable"?

(Sorry for late review)

inexorabletash:
Speaking of late reviews, I noticed that at least some errors are now StructuredSerialize-able. So making success/failure more explicit should be done. Follow-up PRs!

index.bs
abhishek-shanthkumar: Thanks, I'll update this in a follow-up PR.

abhishek-shanthkumar:
> Speaking of late reviews, I noticed that at least some errors are now StructuredSerialize-able. So making success/failure more explicit should be done. Follow-up PRs!

Could you elaborate a bit on what you mean? Were `DOMException`s not StructuredSerialize-able before?

inexorabletash:
> > Speaking of late reviews, I noticed that at least some errors are now StructuredSerialize-able. So making success/failure more explicit should be done. Follow-up PRs!
> 
> Could you elaborate a bit on what you mean? Were `DOMException`s not StructuredSerialize-able before?

Not until 2019. https://github.com/whatwg/webidl/pull/732

Which is a long time ago, but this spec is pretty old. :)

PR #440, opened 2025-02-26
"Convert a value to a key" should return invalid for detached array buffers

### LINKED ISSUE #417: Inconsistency between spec and tests for detached arraybuffers; spec for keys says don't throw, test for keys says throw, spec for values says throw
While addressing a Firefox test failure in [idb-binary-key-detached.htm](https://searchfox.org/mozilla-central/source/testing/web-platform/tests/IndexedDB/idb-binary-key-detached.htm) in https://phabricator.services.mozilla.com/D202947 we became aware of a chromium issue https://issues.chromium.org/issues/40282817 which correctly points out that the IDB invocation of [getting a copy of the bytes held by the buffer source](https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy) in [convert a value to a key](https://w3c.github.io/IndexedDB/#convert-a-value-to-a-key) means we shouldn't throw because webidl says to "return the empty byte sequence".

However, [StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/structured-data.html#structuredserializeinternal) step 13.2.1 says:
> If [IsDetachedBuffer](https://tc39.es/ecma262/#sec-isdetachedbuffer)(value) is true, then throw a ["DataCloneError"](https://webidl.spec.whatwg.org/#datacloneerror) [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException).

Arguably the current test behavior (which Blink implements) is the most consistent and developer-friendly behavior.  It would be weird for a detached arraybuffer to fail as a value but succeed with a potentially incorrect result as a key.

I would propose we:
- Modify the IDB spec to explicitly return invalid if "IsDetachedBuffer".
  - Note this will throw a DataError rather than the TypeError expected by the test.  I'm not opposed to retaining TypeError, but it would be good to have some rationale for choosing that given that we otherwise just seem to return DataError and structured serialization returns DataCloneError.
- Update the existing test to expect DataError
- Add a case for a key that is a JS array that contains a detached arraybuffer.

SteveBeckerMSFT:
Closes #417 by updating ["7.4 Convert a value to a key"](https://www.w3.org/TR/IndexedDB/#convert-value-to-key) to return invalid when the `input` buffer source is detached.

 * [X] Confirmed there are no ReSpec/BikeShed errors or warnings.
 * [X] Modified Web platform tests (https://chromium-review.googlesource.com/c/chromium/src/+/6303007)

Implementation commitment:

 * [ ] WebKit
 * [X] Chromium (https://issues.chromium.org/issues/40282817)
 * [X] Gecko

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/440.html" title="Last updated on Feb 26, 2025, 8:49 PM UTC (d7fc3fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/440/a72426f...d7fc3fc.html" title="Last updated on Feb 26, 2025, 8:49 PM UTC (d7fc3fc)">Diff</a>


index.bs
```diff
@@ -6366,9 +6366,11 @@ steps may throw an exception.
       <!-- Binary -->
       : If |input| is a [=buffer source type=]
       ::
+          1. If [$IsDetachedBuffer$](|input|) is true then return invalid.
```
inexorabletash, line 6369: IsDetachedBuffer() takes an ArrayBuffer or a SharedArrayBuffer, but buffer source type includes the views. I think you'll need to dig the buffer out of the view in those cases?

inexorabletash (changes requested):
Two more things:

* Add yourself to the Acknowledgements section
* Add an entry to the Revision History section

index.bs
SteveBeckerMSFT: Thanks for the review!  I found the new steps from the WebIDL spec here:

https://webidl.spec.whatwg.org/#buffersource-detached

SteveBeckerMSFT: I updated the latest revision to use the shorthand for these steps.

inexorabletash (approved):

inexorabletash:
Thanks @SteveBeckerMSFT !

PR #447, opened 2025-04-11
Address transaction state edge cases

### LINKED ISSUE #436: Wait loop questions
In [Upgrading a database](https://w3c.github.io/IndexedDB/#upgrade-transaction-steps) step 11 is supposed to wait for the transaction to finish.

A [transaction is finished](https://w3c.github.io/IndexedDB/#transaction-finished) once it enters the `finished` state (after `commit` or `abort`)

Calling `indexedDB.open("test", 1)` will cause an upgrade, and then step 10.5 _could_ change the state (via autocommit, or user action), but then we instantly set it back to inactive in step 10.6

* Should step 11 instead be waiting for the task created in step 10 to finish? Or am I missing something here?

Given the following snippet:
```js
let db;
const open_rq = indexedDB.open(undefined, 9);

open_rq.onupgradeneeded = function (e) {
    db = e.target.result;
};
    
open_rq.onsuccess = function (e) {
    db.close();
    let open_rq2 = indexedDB.open(db.name, 10);
};
```

Once we get to [close a database connection](https://w3c.github.io/IndexedDB/#close-a-database-connection) step 3. we will be stuck in an infinite loop, because at no point has the transaction gone to the finished state. (The `onupgradeneeded` callback does nothing that would cause an abort or commit)

* Is there a spec step im missing that handles this case?

inexorabletash:
* Clarify that only inactive transactions should auto-commit.

* Fix upgrade steps to correctly handle aborted transactions.

Resolves #436

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/447.html" title="Last updated on Apr 24, 2025, 9:54 PM UTC (4ed56d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/447/ee73e87...4ed56d0.html" title="Last updated on Apr 24, 2025, 9:54 PM UTC (4ed56d0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -43,7 +43,6 @@ spec: ecma-262; urlPrefix: https://tc39.github.io/ecma262/
 spec: storage; urlPrefix: https://storage.spec.whatwg.org/
     type: dfn
         text: storage bucket; url: storage-bucket
-        text: storage key; url: storage-key
 </pre>
 
 <style>
@@ -1054,7 +1053,7 @@ The <dfn>lifetime</dfn> of a
     stores=] and [=/indexes=].
 
 1. The implementation must attempt to <dfn lt="commit|committed">commit</dfn>
-    a transaction when all [=/requests=] placed against the
+    an [=transaction/inactive=] transaction when all [=/requests=] placed against the
     transaction have completed and their returned results handled, no
     new requests have been placed against the transaction, and the
     transaction has not been [=transaction/aborted=]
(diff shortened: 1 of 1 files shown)


inexorabletash:
@stelar7 can you take a look?

stelar7 (approved):
Would like some more distinct steps that shows where/when the transaction should auto-commit, but this fixes the other part of the issue 👍

SteveBeckerMSFT (approved):
Sorry for the delay.  Looks good to me.  Thanks for fixing the extra brackets around my detached buffer change.  Not sure how I missed that.

PR #453, opened 2025-05-01
Correct record value serialization in cursor iteration

### LINKED ISSUE #452: Possible logic inversion in `iterate a cursor`
[openCursor](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-opencursor) creates a cursor with the [key only flag](https://w3c.github.io/IndexedDB/#cursor-key-only-flag) set to false, source as an ObjectStore and then iterates it.

[iterate a cursor](https://w3c.github.io/IndexedDB/#iterate-a-cursor) step 13 checks if the key only flag is false, and then tries to access the referenced value.

Only Index Records have [referenced values](https://w3c.github.io/IndexedDB/#index-referenced-value)

Should step 13 be checking if the flag is true? If source is an index? Or am I missing something?

SteveBeckerMSFT:
Implements @inexorabletash's suggestion to fix #452.

 * [X ] Confirmed there are no ReSpec/BikeShed errors or warnings.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6019,7 +6019,7 @@ To <dfn>iterate a cursor</dfn> with |targetRealm|, |cursor|, an optional
 
 1. If |cursor|'s [=cursor/key only flag=] is false, then:
 
-    1. Let |serialized| be |found record|'s [=index/referenced value=].
+    1. Let |serialized| be |found record|'s [=/value=] if |source| is an [=/object store=], or |found record|'s [=index/referenced value=] otherwise.
     1. Set |cursor|'s [=cursor/value=] to
         [=ECMAScript/!=] [$StructuredDeserialize$](|serialized|, |targetRealm|)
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6019,7 +6019,7 @@ To <dfn>iterate a cursor</dfn> with |targetRealm|, |cursor|, an optional
 
 1. If |cursor|'s [=cursor/key only flag=] is false, then:
 
-    1. Let |serialized| be |found record|'s [=index/referenced value=].
+    1. Let |serialized| be |found record|'s [=/value=] if |source| is an [=/object store=], or |found record|'s [=index/referenced value=] otherwise.
```
SteveBeckerMSFT, line 6022: Follow up question:  Can this fail?  Do we need error handling in these steps?

inexorabletash (approved):
LGTM!

PR #455, opened 2025-05-05
Add handle to cursor to avoid exposing internal state to script

### LINKED ISSUE #445: asynchronously execute a request source object confusion
[asynchronously execute a request](https://w3c.github.io/IndexedDB/#asynchronously-execute-a-request) takes a source object.

In most places this refers to an `IDBObjectStore`, `IDBIndex`, or `IDBCursor`. (handle objects)
However, in [IDBCursor::continue](https://w3c.github.io/IndexedDB/#dom-idbcursor-continue) step 11, it seems to refer to the "internal" objects (`Object Store`, and `Index`)
Since JS is never exposed to "internal" types, having them there seems very odd.

Is this a typo (`this’s source` -> `this`) for some of the IDBCursor methods? (`advance`, `continue`, `continuePrimaryKey`)

Or should the methods that create an IDBCursor pass a different object instead?
i.e [IDBObjectStore::openCursor](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-opencursor) creates one with `this's object store`, which would be an "internal" object

SteveBeckerMSFT:
Closes #445 by updating cursors to include an object store handle or an index handle.

 * [X] Confirmed there are no ReSpec/BikeShed errors or warnings.
 * [X] Modified Web platform tests (https://chromium-review.googlesource.com/c/chromium/src/+/6514010)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6719,7 +6719,7 @@ For the revision history of the second edition, see [that document's Revision Hi
 * Clarify that only [=transaction/inactive=] [=/transactions=] should attempt to auto-commit. (<#436>)
 * Correct [=/upgrade a database=] steps to handle aborted transactions. (<#436>)
 * Update [=/iterate a cursor=] value serialization to use [=/value=] for [=/object stores=] instead of [=index/referenced values=]. (<#452>)
-* Add [=cursor/source handle=] to [=/cursor=] to avoid exposing internal indexes and object stores to script.
+* Add [=cursor/source handle=] to [=/cursor=] to avoid exposing internal indexes and object stores to script. (<#445>)
 
 <!-- ============================================================ -->
 # Acknowledgements # {#acknowledgements}
--- a/index.bs
+++ b/index.bs
@@ -1380,15 +1380,17 @@ A <dfn>cursor</dfn> is used to iterate over a range of records in an
 
 <div dfn-for=cursor>
 
-A [=cursor=] has a <dfn>transaction</dfn>, the [=/transaction=]
-that was [=transaction/active=] when the cursor was created.
+A [=cursor=] has a <dfn>source handle</dfn>, which is the [=/index handle=] or the [=/object store handle=] that opened the cursor.
+
+A [=cursor=] has a <dfn>transaction</dfn>, which is the [=/transaction=] from the cursor's [=cursor/source handle=].
 
 A [=cursor=] has a <dfn>range</dfn> of records in either an
 [=/index=] or an [=/object store=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1380,15 +1380,17 @@ A <dfn>cursor</dfn> is used to iterate over a range of records in an
...
 A [=cursor=] has a <dfn>range</dfn> of records in either an
 [=/index=] or an [=/object store=].
 
-A [=cursor=] has a <dfn>source</dfn> that indicates which [=/index=]
-or an [=/object store=] is associated with the records over which
-the [=cursor=] is iterating.
+A [=cursor=] has a <dfn>source</dfn>, which is an [=/index=] or an [=/object store=] from the cursor's [=cursor/source handle=].
+The cursor's [=cursor/source=] indicates which [=/index=] or [=/object store=] is associated with the records over which the [=cursor=] is iterating.
+When [=cursor/source handle=] is an [=index handle=], set [=cursor/source=] to [=index-handle/index|the index handle's associated index=].
+Otherwise, set [=cursor/source=] to [=object-store-handle/object store|the object store handle's associated object store=].
```
inexorabletash, line 1393: These lines read like imperative algorithm steps, but they don't have a trigger. How about this:

```suggestion
If the cursor's [=cursor/source handle=] is an [=index handle=], then the cursor's [=cursor/source=] is [=index-handle/index|the index handle's associated index=].
Otherwise, cursor's [=cursor/source=] is [=object-store-handle/object store|the object store handle's associated object store=].
```

index.bs
```diff
@@ -6721,6 +6719,7 @@ For the revision history of the second edition, see [that document's Revision Hi
 * Clarify that only [=transaction/inactive=] [=/transactions=] should attempt to auto-commit. (<#436>)
 * Correct [=/upgrade a database=] steps to handle aborted transactions. (<#436>)
 * Update [=/iterate a cursor=] value serialization to use [=/value=] for [=/object stores=] instead of [=index/referenced values=]. (<#452>)
+* Add [=cursor/source handle=] to [=/cursor=] to avoid exposing internal indexes and object stores to script.
```
inexorabletash, line 6722:
```suggestion
* Add [=cursor/source handle=] to [=/cursor=] to avoid exposing internal indexes and object stores to script. (<#445>)
```

inexorabletash (changes requested):
Overall looks good, with some suggestions.

index.bs
SteveBeckerMSFT: Thanks for the edit.

SteveBeckerMSFT:
Thanks for the review!

inexorabletash (approved):

Issue #433, opened 2024-12-02
Spec clarification regarding implied steps


stelar7:
From working on the [Ladybird](https://ladybird.org/) implementation of this spec, I've encountered some parts that could use clarification..

1. [Opening a database connection](https://w3c.github.io/IndexedDB/#opening) step 10.8 says `If the upgrade transaction was aborted`, but there is no mention elsewhere in the steps where the `upgrade transaction` comes from. (Also not how to check if its aborted.. See next number)

2. The only way to tell if a transaction was aborted seems to be if [the transaction has an error](https://w3c.github.io/IndexedDB/#transaction-error), but when calling [IDBTransaction::abort()](https://w3c.github.io/IndexedDB/#dom-idbtransaction-abort), the error is set to null (It calls [abort a transaction](https://w3c.github.io/IndexedDB/#abort-a-transaction) with a null parameter.). 
So this abort is not marking the transaction as aborted?

3. In [abort a transaction](https://w3c.github.io/IndexedDB/#abort-a-transaction), step 6.3.1, it says to use `the open request associated with transaction`, but there is no spec steps that sets this anywhere

4. In [delete a database](https://w3c.github.io/IndexedDB/#delete-a-database), step 5. wants `the set of all connections associated with db`, but there are no steps to create this association

inexorabletash:
Thanks! This sort of feedback is great. This spec predates the modern style of writing out explicit algorithms, so it's unsurprising that there are still gaps to plug.

> * [Opening a database connection](https://w3c.github.io/IndexedDB/#opening) step 10.8 says `If the upgrade transaction was aborted`, but there is no mention elsewhere in the steps where the `upgrade transaction` comes from. 

Agreed. It looks like an easy fix would be for the "[upgrade a database](https://w3c.github.io/IndexedDB/#upgrade-a-database)" steps to return the transaction.

> * The only way to tell if a transaction was aborted seems to be if [the transaction has an error](https://w3c.github.io/IndexedDB/#transaction-error), but when calling [IDBTransaction::abort()](https://w3c.github.io/IndexedDB/#dom-idbtransaction-abort), the error is set to null (It calls [abort a transaction](https://w3c.github.io/IndexedDB/#abort-a-transaction) with a null parameter).
>   So this abort is not marking the transaction as aborted?

The _intent_ is that an explicit `abort()` does abort the transaction, but the value reported via the [error getter](https://w3c.github.io/IndexedDB/#dom-idbtransaction-error) is null. 

[...]

PR #458, opened 2025-05-22
Check the upgrade request's error instead of the upgrade transaction

stelar7:
Some work towards fixing the issues mentioned in #433 

 * [ ] Confirmed there are no ReSpec/BikeShed errors or warnings.

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/stelar7/IndexedDB/pull/458.html" title="Last updated on Aug 11, 2025, 7:36 PM UTC (9df6a0a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/458/8a573ea...stelar7:9df6a0a.html" title="Last updated on Aug 11, 2025, 7:36 PM UTC (9df6a0a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -947,8 +947,12 @@ which is an [=/event loop=].
 A [=/transaction=] has a <dfn>request list</dfn> of pending [=/requests=]
 which have been made against the transaction.
 
-A [=/transaction=] has a <dfn>error</dfn> which is set if the
-[=/transaction=] is [=transaction/aborted=].
+A [=/transaction=] has a <dfn>error</dfn> which is set if the [=/transaction=] is [=transaction/aborted=].
+
+<div class=note>
+  <span class=marker>NOTE:</span>
+  Implementors need to keep in mind that the value "null" is considered an error, as it is set from {{IDBTransaction/abort()}}
+</div>
 
 A [=/transaction=]'s [=get the parent=] algorithm returns the
 transaction's [=transaction/connection=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4957,14 +4957,13 @@ To <dfn>open a database connection</dfn> with |storageKey| which requested the [
...
 
-    1. Run [=upgrade a database=]
-        using |connection|, |version| and |request|.
+    1. Let |transaction| be the result of running [=upgrade a database=] with |connection|, |version| and |request|.
 
     1. If |connection| was [=connection/closed=],
         return a newly [=exception/created=]
         "{{AbortError}}" {{DOMException}} and abort these steps.
 
-    1. If the [=/upgrade transaction=] was aborted, run the steps
```
evanstade, line 4967: I don't think you need to return a value from `upgrade a database` as the request's transaction will be set to the transaction, although only if it succeeded. If it was aborted, the request's transaction will be null and done flag will be false. So this can change to something like "If |request|'s [=request/transaction=] is null, run the steps to..."

inexorabletash: Thanks for this analysis, @evanstade. Agreed - I think this recommendation is sufficient. Even easier than what I recommended over in https://github.com/w3c/IndexedDB/issues/433#issuecomment-2512330086

SteveBeckerMSFT: Good find @stelar7!  Upgrade transaction completion does [indeed reset the request's transaction to null](https://w3c.github.io/IndexedDB/#ref-for-request-transaction%E2%91%A2).  The [web platform tests also verify](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/IndexedDB/upgrade-transaction-lifecycle-committed.any.js;drc=658153df0cd23ec41cb9f77e2689d342278d0ccd;l=53) this behavior.

Going back to the drawing board, maybe another potential alternative involves using the request's error, which the abort sets?

stelar7: Is there some reason we dont want to have the return? (what i feel is the simple solution)

evanstade: Good catch @stelar7.

I like Steve's idea.

> Is there some reason we dont want to have the return?

In general it seems preferable to work with the state that is already available, rather than making more state available to more of the system (especially when it is redundant). Returning the transaction changes its lifetime. There are fewer hypothetical side effects to reading state that already exists in the request.

stelar7: Updated to use the requests error 👍

SteveBeckerMSFT (approved):

PR #471, opened 2025-08-13
Specify that the requests error should go back to the unset state

stelar7:
While running `asynchronously execute a request`, `request`'s `error` is set to `undefined`.  
In most cases this would mean the JS value `undefined`, but in this case, its supposed to go back to the unset state.

Im not sure if there is a better spec wording for this 😅 

 * [ ] Confirmed there are no ReSpec/BikeShed errors or warnings.

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/471.html" title="Last updated on Aug 13, 2025, 1:47 PM UTC (b6eee7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/471/c59639b...b6eee7d.html" title="Last updated on Aug 13, 2025, 1:47 PM UTC (b6eee7d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5332,7 +5332,7 @@ created [=/request=] belongs to is [=transaction/aborted=] using the steps to
         1. Otherwise:
 
             1. Set |request|'s [=request/result=] to |result|.
-            1. Set |request|'s [=request/error=] to undefined.
+            1. Unset |request|'s [=request/error=].
             1. [=Fire a success event=] at |request|.
 
 1. Return |request|.


SteveBeckerMSFT:
Hi @stelar7, thanks for the change.  Could you please elaborate more on why this change is required?

The language `Set "X" to undefined.` is used throughout the spec.  Are undefined and unset equivalent?

stelar7:
Having the spec say `set` implies that it has a value, but in this case it should go back to the 'no value has been set' state.  

It might just feel that way to me because this is the only field that has the concept of `unset` in combination with an optional value 

I'm fine with closing this if you think the current text clearer in wrt. how the specs are written

SteveBeckerMSFT:
I think the current text is probably OK.  Setting to `undefined` should be equivalent to the `no value has been set` state.

Issue #475, opened 2025-10-04
Explicitly define how to handle `queryOrOptions` being omitted in `getAll`/`getAllKeys`


jfdoming:
The [IDL definition](https://www.w3.org/TR/IndexedDB/#idl-index) for `getAll` and `getAllKeys` lists `queryOrOptions` as `optional`; however, the algorithm [5.12 Creating a request to retrieve multiple items](https://www.w3.org/TR/IndexedDB/#create-request-to-retrieve-multiple-items) doesn't handle the case where `queryOrOptions` is not defined (and the [description](https://www.w3.org/TR/IndexedDB/#dom-idbindex-getallkeys) of the method doesn't either). There _is_ the following note at the bottom of the algorithm:
> NOTE: The _range_ can be a [key](https://www.w3.org/TR/IndexedDB/#key) or [key range](https://www.w3.org/TR/IndexedDB/#key-range) (an [IDBKeyRange](https://www.w3.org/TR/IndexedDB/#idbkeyrange)) identifying the [record](https://www.w3.org/TR/IndexedDB/#object-store-record) items to be retrieved. If null or not given, an [unbounded key range](https://www.w3.org/TR/IndexedDB/#unbounded-key-range) is used. If _count_ is specified and there are more than _count_ records in range, only the first _count_ will be retrieved.

Based on the IDL as well as [MDN](https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex/getAllKeys#query) and [these](https://wpt.live/IndexedDB/idbindex_getAllKeys.any.html) Web Platform Tests, I think the note is meant to apply to `queryOrOptions` (it is not clear to me what `range` refers to here). Is that accurate? And if so, would it be worth updating the 5.12 algorithm to explicitly handle this case as a step?

SteveBeckerMSFT:
Thanks for reporting this issue.  #478 proposed a fix, which we are discussing in the PR.

PR #478, opened 2025-10-15
Adjust how the direction is defaulted in "Creating a request to retrieve multiple items"

stelar7:
Since [getAllKeys](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-getallkeys) takes an `any` parameter, we cant be sure that the object we get is of the type `IDBGetAllOptions`. As such, there is no default values applied to the object passed in.

If the object passed then doesnt contain direction, we will try to call `retrieve_multiple_items_from_an_XXX` with an undefined value for `direction`. This in turn will result in 0 records being returned. I assume this is not the intention, and as such this PR aims to set the default value of `"next"` in those cases.

The same would apply to `query` and `count`, but those are explicitly handled to return an unbounded key, and infinite results in other places.

Ideally there would be some wording like `"Set direction to queryOrOptions['direction'] if it exists."`, but I'm not sure of the correct spec syntax in these cases

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/stelar7/IndexedDB/pull/478.html" title="Last updated on Oct 17, 2025, 8:35 PM UTC (0a4563a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/478/f704918...stelar7:0a4563a.html" title="Last updated on Oct 17, 2025, 8:35 PM UTC (0a4563a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5629,13 +5629,11 @@ To <dfn>create a [=request=] to retrieve multiple items</dfn> from an [=object s
 
 1. Let |range| be a [=key range=].
 
-1. Let |direction| be a [=cursor/direction|cursor direction=].
+1. Let |direction| be "{{IDBCursorDirection/next}}".
 
 1. If running [=is a potentially valid key range=] with |queryOrOptions| is true, then:
 
     1. Set |range| to the result of [=/converting a value to a key range=] with |queryOrOptions|.  Rethrow any exceptions.
-
-    1. Set |direction| to "{{IDBCursorDirection/next}}".
 
 1. Else:
 


SteveBeckerMSFT:
Thanks for identifying this issue and proposing a fix.  #475 also reported this issue.  Maybe we should update "[creating a request to retrieve multiple items](https://w3c.github.io/IndexedDB/#create-a-request-to-retrieve-multiple-items)" to explicitly handle an undefined `queryOrOptions`?

Prior to `getAllRecords()`, the steps for "[convert a value to a key range ](https://w3c.github.io/IndexedDB/#convert-a-value-to-a-key-range)" handled undefined values.  Specifically, step 2, which returns an unbounded key range for an undefined value.

> If value is undefined or is null, then [throw](https://webidl.spec.whatwg.org/#dfn-throw) a "[DataError](https://webidl.spec.whatwg.org/#dataerror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException) if null disallowed flag is true, or return an [unbounded key range](https://w3c.github.io/IndexedDB/#unbounded-key-range) otherwise.

We could update step 8 in "[creating a request to retrieve multiple items](https://w3c.github.io/IndexedDB/#create-a-request-to-retrieve-multiple-items)" to say something like this: 

> If queryOrOptions is undefined or running [is a potentially valid key range](https://w3c.github.io/IndexedDB/#is-a-potentially-valid-key-range) with queryOrOptions is true, then:

Thoughts?

stelar7:
This issue is slightly different than the one in https://github.com/w3c/IndexedDB/issues/475.
In this case, `queryOrOptions` is just missing the `direction` parameter, and as such is not a potentially valid key range.
```json
{
    "query": "a",
    "count": 1
}
```
This takes us to step 9.3, where we encounter an `undefined` value, and its never set to its default

SteveBeckerMSFT:
Thanks for clarifying.  For the dictionary without a direction case, I think the default value in the `IDBGetAllOptions dictionary` is applied:

```
dictionary IDBGetAllOptions {
...
  [IDBCursorDirection](https://w3c.github.io/IndexedDB/#enumdef-idbcursordirection) direction = "next";
};
```
https://w3c.github.io/IndexedDB/#dictdef-idbgetalloptions

See also the WebIDL spec: https://webidl.spec.whatwg.org/#dfn-dictionary-member-default-value

stelar7:
The `IDBGetAllOptions dictionary` has a default yes, but there is afaik nothing that makes the object we pass be of the type of that dictionary. (assuming the `potentially valid key range` check fails)

SteveBeckerMSFT:
Any JavaScript object may become an IDBGetAllOptions dictionary when passed into a function like getAllRecords().  Dictionary types do not exist in JavaScript.  This looks like the relevant section from the webidl spec:

https://webidl.spec.whatwg.org/#js-dictionary

You can also see this in action in Chromium's IDBGetAllOptions implementation, which is automatically generated by the V8 bindings code:

https://source.chromium.org/chromium/chromium/src/+/main:out/chromeos-Debug/gen/third_party/blink/renderer/bindings/modules/v8/v8_idb_get_all_options.cc;drc=240825a2c96b5bd0ae6d124919de323054ac6080;bpv=1;bpt=1;l=32?q=IDBGetAllOptions%20-file:out&ss=chromium%2Fchromium%2Fsrc

stelar7:
> Any JavaScript object may become an IDBGetAllOptions dictionary when passed into a function like getAllRecords().

Since the parameter type for `getAllKeys` is `any`, `queryOrOptions` doesnt know its supposed to be `IDBGetAllOptions`. (Atleast thats my understanding of `any`)

There are no steps that say i.e `Convert 'queryOrOptions' to the 'IDBGetAllOptions' type`.
What about adding a step at the start of step `9. Else:` for that?

SteveBeckerMSFT:
I don't think a conversion step is needed.  

Any JavaScript argument that is an 'object' type may serve as an `IDBGetAllOptions` dictionary.  For `getAll()` and friends, when passed an argument that is an 'object' type and is not a type of IDB key, the browser must attempt to get the values from the object's keys using 'query', 'count' and 'direction'.  When the resulting values are undefined, the browser must use the default values defined by the [`IDBGetAllOptions` dictionary](https://www.w3.org/TR/IndexedDB/#dictdef-idbgetalloptions).

It's unfortunate that `getAll()` and friends use `any`, which forces the browser to differentiate between object types that might be a valid IDB key, which includes Dates, Arrays and ArrayBuffers, and object types that might serve as the `IDBGetAllOptions` dictionary.  To maintain backward compat for `getAll()` and friends, the browser must continue using Dates, Arrays and ArrayBuffers as IDBKeyRanges, even if they are invalid IDB keys (i.e. a NaN Date).  We don't want to change existing behavior for any IDB key types.

See "[is a potentially valid key range](https://www.w3.org/TR/IndexedDB/#is-a-potentially-valid-key-range)", which tries to handle these cases.

stelar7:
> When the resulting values are undefined, the browser must use the default values defined by the [`IDBGetAllOptions` dictionary](https://www.w3.org/TR/IndexedDB/#dictdef-idbgetalloptions).

Where is this specified? I cant find anything anywhere that says to use those defaults.. 😅

PR #479, opened 2025-10-25
Mark transactions inactive during key serialization

### LINKED ISSUE #476: Transactions should be marked as inactive during key serialization
Chromium ([bug](https://issues.chromium.org/issues/40050469)), Gecko ([bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1598164)), and WebKit ([bug](https://bugs.webkit.org/show_bug.cgi?id=227773)) all make transactions inactive during structured serialization of object values. The spec addresses this in the ["clone a value"](https://w3c.github.io/IndexedDB/#clone) algorithm (see also [`structured-clone-transaction-state.any.js`](https://github.com/web-platform-tests/wpt/blob/4ae2deda519dba8c9a13df34a710b66e03446f57/IndexedDB/structured-clone-transaction-state.any.js)).

However, the spec does not consider _keys_, which can be arbitrarily set using the second argument to `put()`/`add()`:

```js
const array = [0]
Object.defineProperty(array, '0', {
  get() { 
    // can continue to call `store.put` here
    return 0
  }
})
store.put({}, array)
```

This unfortunately leads to a discrepancy in browser behavior, as demonstrated by this minimal repro:

```html
<!doctype html>
<script>    
const request = indexedDB.open(Math.random(), 1)

request.onupgradeneeded = (e) => {
  const db = e.target.result
  db.createObjectStore('store')
}

request.onsuccess = (e) => {
  const db = e.target.result
  const store = db.transaction('store', 'readwrite')
    .objectStore('store')
  
  const array = [0]
  Object.defineProperty(array, '0', { 
    get() {
      store.put({}, [1])
      return 0
    }
  })

  try {
    store.put({}, array)
  } catch (err) {
    console.log('Error', err)
  }
}
</script>
```

Results:

- Chromium 140: `put`s succeed
- GNOME Web 46.5 (WebKit 605.1.15): `puts` succeed
- Firefox 143: throws `DOMException: A request was placed against a transaction which is currently not active, or which is finished.`

(Note that recursive calls result in a `Maximum call stack size exceeded` error in both Chromium and WebKit.)

My intuition is that the spec should probably match Firefox's behavior. It seems sensible for both values and keys to throw because the transaction is not currently active.

nolanlawson:
Closes #476 

 * [x] Modified Web platform tests (https://github.com/web-platform-tests/wpt/pull/55660)

Implementation commitment:

 * [ ] Gecko (Already passes 2/5 tests) (https://bugzilla.mozilla.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/pull/479.html" title="Last updated on Nov 9, 2025, 10:38 PM UTC (b82c2db)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IndexedDB/479/f704918...b82c2db.html" title="Last updated on Nov 9, 2025, 10:38 PM UTC (b82c2db)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3010,12 +3010,24 @@ To <dfn>add or put</dfn> with |handle|, |value|, |key|, and |no-overwrite flag|,
 
 1. If |key| was given, then:
 
+    1. [=/Assert=]: |transaction|'s [=transaction/state=] is [=transaction/active=].
+
+    1. Set |transaction|'s [=transaction/state=] to [=transaction/inactive=].
+
+          NOTE:
+            The [=/transaction=] is made [=transaction/inactive=] so that getters or other side effects triggered by the cloning operation are unable to make additional requests against the transaction.
+
     1. Let |r| be the result of [=/converting a
         value to a key=] with |key|. Rethrow any exceptions.
 
+    1. Set |transaction|'s [=transaction/state=] to [=transaction/active=].
+
     1. If |r| is "invalid value" or "invalid type", [=exception/throw=] a "{{DataError}}" {{DOMException}}.
 
+    1. Let |r| is "invalid value" or "invalid type", [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+
     1. Let |key| be |r|.
+
 
 1. Let |targetRealm| be a user-agent defined [=ECMAScript/Realm=].
 
--- a/index.bs
+++ b/index.bs
@@ -3010,8 +3010,7 @@ To <dfn>add or put</dfn> with |handle|, |value|, |key|, and |no-overwrite flag|,
 
 1. If |key| was given, then:
 
-    1. Let |r| be the result of [=/converting a value to a key during a
-        transaction=] with |key| and |transaction|. Rethrow any exceptions.
+    1. Let |r| be the result of [=/converting a value to a key=] with |key| and |transaction|. Rethrow any exceptions.
 
     1. If |r| is "invalid value" or "invalid type", [=exception/throw=] a "{{DataError}}" {{DOMException}}.
 
@@ -4511,7 +4510,7 @@ The <dfn method for=IDBCursor>continue(|key|)</dfn> method steps are:
 1. If |key| is given, then:
 
     1. Let |r| be the result of [=/converting a
-        value to a key during a transaction=] with |key| and |transaction|. Rethrow any exceptions.
+        value to a key=] with |key| and |transaction|. Rethrow any exceptions.
 
     1. If |r| is "invalid value" or "invalid type", [=exception/throw=] a "{{DataError}}" {{DOMException}}.
 
@@ -4570,14 +4569,14 @@ The <dfn method for=IDBCursor>continuePrimaryKey(|key|, |primaryKey|)</dfn> meth
     [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
 
 1. Let |r| be the result of [=/converting a value to
-    a key during a transaction=] with |key| and |transaction|. Rethrow any exceptions.
+    a key=] with |key| and |transaction|. Rethrow any exceptions.
 
 1. If |r| is "invalid value" or "invalid type", [=exception/throw=] a "{{DataError}}" {{DOMException}}.
 
 1. Let |key| be |r|.
 
 1. Let |r| be the result of [=/converting a value
-     to a key during a transaction=] with |primaryKey| and |transaction|. Rethrow any exceptions.
+     to a key=] with |primaryKey| and |transaction|. Rethrow any exceptions.
 
 1. If |r| is "invalid value" or "invalid type", [=exception/throw=] a "{{DataError}}" {{DOMException}}.
 
@@ -6481,8 +6480,8 @@ The steps return an ECMAScript value.
 
 <div algorithm>
 
-To <dfn>convert a value to a key</dfn> with an ECMAScript value |input|, and an optional
-[=/set=] |seen|, run the following steps.
+To <dfn lt="recursively convert a value to a key|recursively converting a value to a key">recursively convert a value to a key</dfn>
+with an ECMAScript value |input|, and an optional [=/set=] |seen|, run the following steps.
 The result of these steps is a [=/key=], or "invalid value", or "invalid type", or the
 steps may throw an exception.
 
@@ -6549,7 +6548,7 @@ steps may throw an exception.
               1. Let |entry| be [=ECMAScript/?=] [$Get$](|input|, |index|).
 
               1. Let |key| be the result of
-                  [=/converting a value to a key=] with arguments |entry|
+                  [=/recursively converting a value to a key=] with arguments |entry|
                   and |seen|.
 
               1. [=ECMAScript/ReturnIfAbrupt=](|key|).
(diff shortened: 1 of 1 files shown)


nolanlawson:
Note I only added the minimal logic to mark the transaction inactive when a key is provided to `IDBObjectStore#add`/`put`. I did not:

- mark the transaction inactive anywhere else that we convert a key to a value, e.g. `indexedDB.cmp` or `IDBCursor#continue`, because this matches Firefox's implementation (and Chromium's as well – both browsers just throw `InternalError: too much recursion` / `RangeError: Maximum call stack size exceeded`
- refactor to reuse the existing logic from [`clone`](file:///home/nolan/workspace/IndexedDB/index.html#clone) – this seemed like overkill since the logic was only needed in one place.

asutherland:
Thanks for calling out other places where the spec converts a key to a value!  (And for filing the spec issue in the first place!)

For places like indexedDB.cmp and the IDBKeyRange methods (only/lowerBound/upperBound/bound/includes) where we don't have a transaction context to mark inactive and there's no complex state management going on it seems fine to not change the logic.

But for [continue](https://w3c.github.io/IndexedDB/#dom-idbcursor-continue) and [continuePrimaryKey](https://w3c.github.io/IndexedDB/#dom-idbcursor-continueprimarykey) we do explicitly have transactions and we do check that they are active and I think there is enough potentially interesting state management going on that it makes sense to consistently apply the same rationale we're using for add/put.  In particular, it seems like we could be in an upgrade transaction when calling one of the continue methods and then nefarious code could attempt to delete the object store or index that the cursor is against with that method on the stack.  I do think it's much less likely for code to be written in a way that could cause a security bug in this situation, but I do think it still applies that there's no reasonable use-case for content to be doing tricky things here and I do think it also simplifies things from a spec perspective.

For example, step 3 of both continue methods is to throw if the underlying source/object store has been deleted.  If we don't mark the transaction as inactive, then [deleteObjectStore](https://w3c.github.io/IndexedDB/#dom-idbdatabase-deleteobjectstore) currently is defined to synchronously delete the store; we "destroy store" without ever going "in parallel" or doing async hand-waving.  [deleteIndex](https://w3c.github.io/IndexedDB/#dom-idbobjectstore-deleteindex) also synchronously does "destroy index" although the 2nd para afterwards does do some hand-waving: "Although this method does not return an [IDBRequest](https://w3c.github.io/IndexedDB/#idbrequest) object, the index destruction itself is processed as an asynchronous request within the [upgrade transaction](https://w3c.github.io/IndexedDB/#upgrade-transaction)."  So we would side-step these edge-cases where the index/objectStore could conceptually be deleted partway through the algorithm and any need to test/specify that by making the transaction inactive during the key conversions.

nolanlawson:
Thanks for the feedback @asutherland! I just checked, and it appears that none of Firefox/Chromium/WebKit actually throw a `TransactionInactiveError` in the case of `continue`/`continuePrimaryKey` (see test draft below).

<details><summary>Click to see test draft</summary>

```js
promise_test(async testCase => {
  const db = await createDatabase(testCase, database => {
    database.createObjectStore('store');
  });

  const transaction = db.transaction(['store'], 'readwrite');
  const objectStore = transaction.objectStore('store');

  objectStore.put({}, 0);
  objectStore.put({}, 1);
  const cursor = await new Promise((resolve, reject) => {
    const cursorReq = objectStore.openCursor();
    cursorReq.onerror = reject;
    cursorReq.onsuccess = e => resolve(e.target.result);
  });

  let getterCalled = false;
  const activeKey = ['value that should not be used'];
  Object.defineProperty(activeKey, '0', {
    enumerable: true,
    get: testCase.step_func(() => {
      getterCalled = true;
      assert_throws_dom('TransactionInactiveError', () => {
        objectStore.get('key');
      }, 'transaction should not be active during key serialization');
      return 'value that should not be used';
    }),
  });
  cursor.continue(activeKey);
  await promiseForTransaction(testCase, transaction);
  db.close();

  assert_true(getterCalled,
    "activeKey's getter should be called during test");
}, 'Transaction inactive during key serialization in IDBCursor.continue()');

promise_test(async testCase => {
  const db = await createDatabase(testCase, database => {
    const objectStore = database.createObjectStore('store');
    objectStore.createIndex('idx', 'name');
  });

  const transaction = db.transaction(['store'], 'readwrite');
  const objectStore = transaction.objectStore('store');

  objectStore.put({ name: 'a' }, 0);
  objectStore.put({ name: 'b' }, 1);
  const idx = objectStore.index('idx')
  const cursor = await new Promise((resolve, reject) => {
    const cursorReq = idx.openCursor();
    cursorReq.onerror = reject;
    cursorReq.onsuccess = e => resolve(e.target.result);
  });

  let getterCalled = false;
  const activeKey = ['value that should not be used'];
  Object.defineProperty(activeKey, '0', {
    enumerable: true,
    get: testCase.step_func(() => {
      getterCalled = true;
      assert_throws_dom('TransactionInactiveError', () => {
        objectStore.get('key');
      }, 'transaction should not be active during key serialization');
      return 'value that should not be used';
    }),
  });
  cursor.continuePrimaryKey(activeKey, 0);
  await promiseForTransaction(testCase, transaction);
  db.close();

  assert_true(getterCalled,
    "activeKey's getter should be called during test");
}, 'Transaction inactive during key serialization in IDBCursor.continuePrimaryKey()');
```

</details>

Firefox does throw for `put`/`add`, though, which is why I originally scoped the fix to just that. Are you proposing that we handle `continue`/`continuePrimaryKey` in the spec despite Firefox's current behavior?

nolanlawson:
After re-reading your comment, I realized you weren't talking specifically about Firefox's implementation. I agree we can make this work with `IDBCursor#continue` / `continuePrimaryKey` as well. Updated the spec language and updated the tests at https://github.com/web-platform-tests/wpt/pull/55660 .

asutherland:
> After re-reading your comment, I realized you weren't talking specifically about Firefox's implementation.

Yes.

> I agree we can make this work with `IDBCursor#continue` / `continuePrimaryKey` as well. Updated the spec language and updated the tests at [web-platform-tests/wpt#55660](https://github.com/web-platform-tests/wpt/pull/55660) .

Thank you; I like the introduction of the wrapper, this seems very clean!

@evanstade and @SteveBeckerMSFT in https://github.com/w3c/IndexedDB/issues/476 I think we were initially only talking about matching Firefox; are you okay with the (consistent) expansion to also cover the cursor continue methods?

evanstade:
There are a lot of methods that take a key as an argument and are associated with a transaction --- many of the ones in `ObjectStore`, not just `put` and `add`. Why are those safe as-is if we think `IDBCursor#continue` should be hardened?

Just based on the name alone, `converting a value to a key during a transaction` sounds like something that should universally be used over `converting a value to a key` if there is an associated transaction. Otherwise we should come up with a name for it that better distinguishes when it should be referenced in the spec.

asutherland:
Yes, I agree we should expand the mechanism to cover [convert a value to a key range](https://w3c.github.io/IndexedDB/#convert-a-value-to-a-key-range) too if you're on board since the same rationale applies.  I somehow had a lot of tunnel vision going on and ignored that algorithm and its many uses.

evanstade:
Yeah it makes sense to apply the change more broadly.

(Sorry for slow reply.)

nolanlawson:
Sorry for the delay. I've made the following changes:

- Renamed the existing algorithm to "_recursively_ convert a value to a key" and made the new "convert a value to a key" take an optional `transaction`. ("Recursively..." is not an amazing name, and I'm open to suggestions. :slightly_smiling_face:)
  - Note: I found this tricky to refactor given the necessary try/catch logic and recursion in the existing algorithm; I hope it's okay to have two separate algorithms.
- Refactored "is a potentially valid key range" and "convert a value to a key range" to pass in the `transaction`.
- The `IDBKeyRange` methods (e.g. `only`/`lowerBound`) do not pass `transaction` since they aren't explicitly associated with a transaction. Same for `indexedDB.cmp()`.
- Similarly, "extract a key from a value using a key path" does not pass `transaction` since there is no explicit transaction, and plus its inputs are already "sanitized" in the sense that they are derived from the store's key path rather than arbitrary user input.

The WPT tests will be a bit tricky to cover all these new scenarios, so I wanted to check if the spec changes look good first. (For example, "Creating a request to retrieve multiple items" calls the new transaction-checking logic multiple times, so you could imagine writing a test where the getter uses a counter to only trigger the `TransactionInactiveError` on the 1st, 2nd, or 3rd time.)

nolanlawson:
Friendly ping @asutherland @evanstade :slightly_smiling_face:

SteveBeckerMSFT:
@nolanlawson thanks for the PR!  I'll also review this soon.

asutherland (approved):
Thank you so much for addressing this important spec problem and my apologies about the delays in my reviewing!

Restating:
- The existing "convert a value to a key" algorithm is renamed to "recursively convert a value to a key" and we introduce a new wrapper "convert a value to a key" that takes the optional transaction and wraps the underlying call to "recursively converting a value to a key" with marking the transaction inactive during the call and restoring it on the way out, making sure we conceptually are catching the exception and re-throwing it on the way out.

@SteveBeckerMSFT it sounded like you also wanted to take a review pass too?

nolanlawson:
@asutherland Thanks for the view! Yes your restating is exactly what I had in mind. :slightly_smiling_face:

